PR #23, opened 2018-02-20
Fix login request

MrYarin:
This will fix the login request bug.


src/Rhisis.Login/LoginHandler.cs
```diff
@@ -52,8 +52,8 @@ public static void OnLogin(LoginClient client, NetPacketBase packet)
                     client.Disconnect();
                     return;
                 }
-
-                LoginPacketFactory.SendServerList(client, user.Username, client.ClustersConnected);
+                Logger.Info(certify.Username);
```
Eastrall: You can remove this log line that is useless now. 😄

PR #43, opened 2018-03-02
[Feature] StatisticsSystem

Steve-Nzr:
This feature is primary based on providing a basic Stats System (STR, STA, DEX, INT & Stats Points).

What has been done here : 
- [Handler] Handling "Upgrade points (ModifyStatus)" packet ▶️ 
- [Packet & Handler] Treat it and dispatch into the newly created StatisticsSystem
- [System] Updating the player in memory
- [WorldPacketFactory] Sending back a confirmation to client ◀️

### DIFF
--- a/src/Rhisis.Core/Network/Packets/World/ModifyStatusPacket.cs
+++ b/src/Rhisis.Core/Network/Packets/World/ModifyStatusPacket.cs
@@ -7,20 +7,20 @@ namespace Rhisis.Core.Network.Packets.World
 {
     public class ModifyStatusPacket
     {
-        public readonly int StrenghtCount;
+        public readonly ushort StrenghtCount;
 
-        public readonly int StaminaCount;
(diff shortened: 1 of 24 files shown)


src/Rhisis.Core/Network/Packets/World/ModifyStatusPacket.cs
```diff
@@ -0,0 +1,26 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Text;
+using Ether.Network.Packets;
+
+namespace Rhisis.Core.Network.Packets.World
+{
+    public class ModifyStatusPacket
+    {
+        public readonly ushort StrenghtCount;
```
Eastrall: You can create readonly properties instrad of readonly variables. Something like: `public ushort StrenghtCount { get; }`

Eastrall:
Nice PR, let's merge! 😄

PR #46, opened 2018-03-04
Add CreateAccountWindow on Installer

Eastrall:
- Add `CreateAccountWindow` design
- Add `CreateAccountViewModel` business logic


Steve-Nzr:
Do we care about coday warning ?

Eastrall:
I guess we can skip them.

PR #49, opened 2018-03-13
Refactoring of ChatSystemEventArgs

Eastrall:
Small refactoring of the ChatSystem and add the `SystemEventArgs` class.


Steve-Nzr:
Appreciate this, more abstract is greater. Let’s see how to improve it with other systems !

Steve-Nzr:
What about Codacy?

Eastrall:
I'm gonna fix the two codacy warnings. In fact, the `IndexOutOfRangeException` should not be thrown by us. I should throw another exception instead. Maybe a `ArgumentException`.
As for the comment, I'll remove it too.

PR #63, opened 2018-03-31
Add dialog data structure and loading

Eastrall:
This PR adds the dialog structure and the loading process in the WorldServer.

Every dialogs are JSON files structured as followed:

```json
{
    "name": "...",
    "oralText": "oral text of the NPC",
    "introText": "Introduction text of the NPC. The first text that appears when you open a dialog.",
    "byeText": "The text displayed when the player closes the dialog box.",
     "links": [
          {
               "id": "Link Id",
               "title": "Title of the link",
               "text": "Text displayed when the player clicks on the link",
          }
     ]
}
```

### DIFF
--- a/bin/data/dialogs/en/MaFl_Lui.json
+++ b/bin/data/dialogs/en/MaFl_Lui.json
@@ -1,0 +1,22 @@
+//
+// MaFl_Lui dialog
+//
+
+{
+	"name": "MaFl_Lui",
+	"oralText": "Come one, come all! Get your supplies here! You're gonna need it!",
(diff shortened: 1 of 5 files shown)


johmarjac:
Does it make sense to have the different languages in one file instead of a separate?

Eastrall:
Actually, you have X folders corresponding to the languages ("en", "fr", "de", ect...). Inside these folders you find all dialog files corresponding to the language.
Just like the shop files, you can have one or more dialogs in one file. For exemple, we could regroup the dialogs of Flaris in the `Flaris.json` same goes for StMorning or Darkon.

johmarjac:
Allright :)

PR #66, opened 2018-04-07
Add /getgold Chat Command

Freezeraid:
type /getgold parameter (gold number)
only one parameter can be used.


Eastrall:
Nice feature! Let's merge!

PR #68, opened 2018-04-10
Add Teleport Chat command + Nav teleport

Freezeraid:


src/Rhisis.World/Game/Chat/TeleportCommand.cs
```diff
@@ -0,0 +1,67 @@
...
+    {
+        [ChatCommand("/tp")]
+        [ChatCommand("/teleport")]
+        public static void TeleportCommand(IPlayerEntity player, string[] parameters)
+        {
+            Logger.Debug("{0} want to teleport", player.Object.Name);
+
+            if (parameters.Length >= 2 && parameters.Length <= 4)
+            {
+                float.TryParse(parameters[0], out float firstValue);
```
Steve-Nzr: Better naming for variables

src/Rhisis.World/Game/Chat/TeleportCommand.cs
```diff
@@ -0,0 +1,67 @@
...
+using Rhisis.World.Game.Entities;
+using Rhisis.World.Packets;
+
+namespace Rhisis.World.Game.Chat
+{
+    public static class TeleportationCommand
+    {
+        [ChatCommand("/tp")]
+        [ChatCommand("/teleport")]
+        public static void TeleportCommand(IPlayerEntity player, string[] parameters)
```
Steve-Nzr: Reduce function size (Local functions) ?

src/Rhisis.World/Game/Chat/TeleportCommand.cs
```diff
@@ -0,0 +1,73 @@
...
+                    Logger.Error("Chat: /teleport command must have 2, 3 or 4 parameters.");
+                    return;
+            }
+        }
+    }
+    public static class TeleportationParameters
+    {
+        public static void TeleportCommandTwoParam(IPlayerEntity player, string[] parameters)
+        {
+            float.TryParse(parameters[0], out float posXValue);
```
Steve-Nzr: Use "is" evaluation keyword and return to make a better error handling 😉

src/Rhisis.World/Game/Chat/TeleportCommand.cs
```diff
@@ -0,0 +1,73 @@
...
+        public static void TeleportCommandThreeParam(IPlayerEntity player, string[] parameters)
+        {
+            int.TryParse(parameters[0], out int mapIdValue);
+            float.TryParse(parameters[1], out float posXValue);
+            float.TryParse(parameters[2], out float posZValue);
+            if (!WorldServer.Maps.TryGetValue(mapIdValue, out Map mapIdResult))
+            {
+                Logger.Error("Cannot find map Id in define files: {0}. Please check you defineWorld.h file.",
+                    mapIdValue);
+            }
```
Steve-Nzr: Return here instead of making an "else" statement.

src/Rhisis.World/Game/Chat/TeleportCommand.cs
```diff
@@ -0,0 +1,73 @@
...
+                case 3: // when you write 3 parameters in the command
+                    TeleportationParameters.TeleportCommandThreeParam(player, parameters);
+                    return;
+                case 4: // when you write 4 parameters in the command
+                    TeleportationParameters.TeleportCommandFourParam(player, parameters);
+                    return;
+                default: // when you write more than 4 or less than 2 parameters in the command
+                    Logger.Error("Chat: /teleport command must have 2, 3 or 4 parameters.");
+                    return;
+            }
```
Steve-Nzr: Send TeleportationPacket here instead of repeating operation each times.

src/Rhisis.World/Game/Chat/TeleportCommand.cs
Eastrall: You should check the return value of `TryParse()` methods. It returns if the parse is successful or not.
Let's imagine you send this command: `/teleport 1 text`. It will parse the first parameter correctly, so `posXValue` value will be equal to 1. But when parsing the second value, since it's plain text, the parsing will fail and then `posZValue` will be equal to 0. That's an error, but you will still teleport the player to X = 1 and Z = 0.

That's why you need to check all the parsed values like:
```cs
if (!float.TryParse(parameters[0], out float posXValue) || !float.TryParse(...))
{
     throw new InvalidArgumentException();
}
```

Eastrall: You should check the return value of `TryParse()` methods. It returns if the parse is successful or not.
Let's imagine you send this command: `/teleport 1 text`. It will parse the first parameter correctly, so `posXValue` value will be equal to 1. But when parsing the second value, since it's plain text, the parsing will fail and then `posZValue` will be equal to 0. That's an error, but you will still teleport the player to X = 1 and Z = 0.

That's why you need to check all the parsed values like:
```cs
if (!float.TryParse(parameters[0], out float posXValue) || !float.TryParse(...))
{
     throw new InvalidArgumentException();
}
```

Eastrall:
Looks good to me. Let's merge!

PR #69, opened 2018-04-12
Feature/trade system

Steve-Nzr:
This PR introduces a complete **trade system**.

It also provides bugs corrections such as :
- dotnet restore now correctly done
- Inventory saving and item deletion


src/Rhisis.World/Systems/Trade/TradeSystem.cs
```diff
@@ -0,0 +1,537 @@
...
+        /// Get the specified entity from the player's context
+        /// Throw if cannot find it
+        /// </summary>
+        /// <exception cref="Exception"></exception>
+        /// <param name="player"></param>
+        /// <param name="id"></param>
+        /// <returns></returns>
+        private static IPlayerEntity GetEntityFromContextOf(IPlayerEntity player, int id)
+        {
+            if (player.Context.FindEntity<IPlayerEntity>(id) is IPlayerEntity entity)
```
Eastrall: `FindEntity<T>` tries to find the entity. If it finds it, it will try to cast it into the type passed as generic. In this case the `IPlayerEntity` type. If it doesn't finds an entity with that id, it will return a `null`.
You can avoid the cast in the `if` statement and do something like: `return player.Context.FindEntity<IPlayerEntity>(id) ?? throw new Exception(...);`

src/Rhisis.World/Systems/Trade/TradeSystem.cs
```diff
@@ -0,0 +1,537 @@
...
+        /// </summary>
+        /// <param name="player"></param>
+        /// <param name="target"></param>
+        /// <returns></returns>
+        private static TradeComponent.TradeConfirm CheckTradeGold(IPlayerEntity player, IPlayerEntity target)
+        {
+            var playerTradeGold = player.Trade.Gold;
+            var targetTradeGold = target.Trade.Gold;
+            var playerGold = player.PlayerData.Gold;
+            var targetGold = player.PlayerData.Gold;
```
Eastrall: do you mean `var targetGold = target.PlayerData.Gold` ?

src/Rhisis.World/Systems/Trade/TradeSystem.cs
```diff
@@ -0,0 +1,537 @@
...
+
+            if (player.PlayerData.Gold < playerTradeGold)
```
Eastrall: You can use `playerGold` instead of `player.PlayerData.Gold`

src/Rhisis.World/Systems/Trade/TradeSystem.cs
```diff
@@ -0,0 +1,537 @@
...
+                return TradeComponent.TradeConfirm.Error;
+
+            if (target.PlayerData.Gold < targetTradeGold)
+                return TradeComponent.TradeConfirm.Error;
+
+            if (targetGold + playerTradeGold < 0)
+                return TradeComponent.TradeConfirm.Error;
+
+            if (playerGold + targetTradeGold < 0)
```
Eastrall: Maybe we can merge all of the conditions `if (player.PlayerData.Gold < playerTradeGold || target.PlayerData.Gold < targetTradeGold || ...) return TradeComponent.TradeConfirm.Error;`

src/Rhisis.Core/Network/Packets/World/Trade/TradeCancelPacket.cs
```diff
@@ -7,7 +7,7 @@ namespace Rhisis.Core.Network.Packets.World.Trade
 {
     public class TradeCancelPacket
     {
-        public int Mode { get; private set; }
+        public readonly int Mode;
```
Eastrall: Keep this as a property by doing : `public int Mode { get; }`

src/Rhisis.World/Game/Components/ItemContainerComponent.cs
```diff
@@ -84,7 +84,11 @@ public int GetItemCount()
         /// </summary>
         /// <param name="slot"></param>
         /// <returns></returns>
-        public Item GetItemBySlot(int slot) => this.Items[slot];
+        public Item this[int slot]
+        {
+            get => Items[slot];
+            set => Items[0] = value;
```
Eastrall: I guess you mean: `set => Items[slot] = value;` ? 😄

src/Rhisis.World/Game/Components/TradeComponent.cs
```diff
@@ -30,5 +31,16 @@ public TradeComponent()
             ItemCount = 0;
             Gold = 0;
         }
+
+        /// <summary>
+        /// Gets the item at a specific slot
+        /// </summary>
+        /// <param name="slot"></param>
+        /// <returns></returns>
+        public Item this[int slot]
```
Eastrall: Not sure you need a indexer proprety in the `TradeComponent`. 
In your component, you have the `ItemContainerComponent` that already has an indexer property. So you can call: `TradeComponent.Items[slot]` instead.

Eastrall:
All good! Merging...

src/Rhisis.World/Game/Components/TradeComponent.cs
```diff
@@ -0,0 +1,35 @@
...
+        public int Gold { get; set; }
+
+        /// <summary>
+        /// Exchange state/progress
+        /// </summary>
+        public TradeState State { get; set; }
+
+        public TradeComponent()
+        {
+            this.TargetId = 0;
```
Skeatwin: Why 0 and not -1 ?
-1 represent null and 0 can be a value, isn't it ?

Issue #71, opened 2018-04-18
Refactoring of system exceptions


Steve-Nzr:
The purpose of this proposal is to introduce our own  ```Exceptions``` with all of our systems.

The idea is to have a **custom error handling** attached to systems, to make it easier to locate errors when exceptions come.

A ```RhisisSystemException``` is thrown in case of the EventArg, the player state, or the desired behavior cannot be performed in the correct way.

e.g : Starting a Trade with yourself, Attacking a player that is too far...

Here's how it actually looks like : 
```csharp
    /// <inheritdoc />
    /// <summary>
    /// Represents a Rhisis exception in an ECS System
    /// </summary>
    public class RhisisSystemException : RhisisException
    {
        /// <inheritdoc />
        /// <summary>
        /// Creates a new <see cref="T:Rhisis.Core.Exceptions.SystemException" /> with a specific error message.
        /// </summary>
        /// <param name="message">Error message</param>
        public RhisisSystemException(string message) :
            base(message)
        {
        }

        /// <inheritdoc />
        /// <summary>
        /// Creates a new <see cref="T:Rhisis.Core.Exceptions.SystemException" /> with a specific error message and a reference to the inner exception that throwed this exception.
        /// </summary>
        /// <param name="message">Error message</param>
        /// <param name="innerException">Inner exception</param>
        public RhisisSystemException(string message, Exception innerException) :
            base(message, innerException)
        {
        }
    }
```

Eastrall:
Solved in PR #72

PR #72, opened 2018-04-23
Refacto/system event

Steve-Nzr:
In this pull request, I worked on **EventArgs** & **systems**.

We're now using **pattern matching** in order to recognize EventArgs.
Il allows us to create our own more easily.


src/Rhisis.World/Systems/NpcShop/NpcShopSystem.cs
```diff
@@ -123,41 +109,41 @@ private void BuyItem(IPlayerEntity player, NpcShopEventArgs e)
...
             }
 
-            var currentTab = npcData.Shop.Items[npcEvent.Tab];
+            var currentTab = npcData.Shop.Items[e.Tab];
 
-            if (npcEvent.Slot < 0 || npcEvent.Slot > currentTab.Count - 1)
+            if (e.Slot < 0 || e.Slot > currentTab.Count - 1)
             {
-                Logger.Error($"ShopSystem: Item slot index was out of tab bounds. Slot: {npcEvent.Slot}");
+                Logger.Error($"ShopSystem: Item slot index was out of tab bounds. Slot: {e.Slot}");
```
Eastrall: Maybe we should `throw` and exception in the cases we didn't send any packet to the client. Or, we can send the packet first and then `throw` the exception.

src/Rhisis.World/Systems/NpcShop/NpcShopSystem.cs
```diff
@@ -123,41 +109,41 @@ private void BuyItem(IPlayerEntity player, NpcShopEventArgs e)
...
                 Logger.Error($"ShopSystem: Shop item id doens't match the item id that {player.Object.Name} is trying to buy.");
                 return;
             }
 
-            if (player.PlayerData.Gold < npcEvent.ItemData.Cost)
+            if (player.PlayerData.Gold < e.ItemData.Cost)
             {
-                Logger.Info($"ShopSystem: {player.Object.Name} doens't have enough gold to buy item {npcEvent.ItemData.Name} at {npcEvent.ItemData.Cost}.");
+                Logger.Info($"ShopSystem: {player.Object.Name} doens't have enough gold to buy item {e.ItemData.Name} at {e.ItemData.Cost}.");
                 WorldPacketFactory.SendDefinedText(player, DefineText.TID_GAME_LACKMONEY);
```
Eastrall: Same comment as before

src/Rhisis.World/Systems/NpcShop/NpcShopSystem.cs
Steve-Nzr: This is part of another issue : https://github.com/Eastrall/Rhisis/issues/71

But yeah, it will be done.

Eastrall:
Nice PR. Let's merge!

PR #83, opened 2018-07-31
Add follow system

Eastrall:
Implements the Follow system. Only for players. Just call the `FollowSystem` with the `FollowEventArgs` to follow another player.

```cs
var followEvent = new FollowEventArgs(targetObjectId, distance);

client.Player.NotifySystem<FollowSystem>(followEvent);
```

### DIFF
--- a/src/Rhisis.World/Systems/BehaviorSystem.cs
+++ b/src/Rhisis.World/Systems/BehaviorSystem.cs
@@ -37,6 +37,9 @@ namespace Rhisis.World.Systems
                     case INpcEntity npc:
                         npc.Behavior.Update(npc);
                         break;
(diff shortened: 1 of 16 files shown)


src/Rhisis.World/Systems/BehaviorSystem.cs
```diff
@@ -36,6 +36,9 @@ public override void Execute(IEntity entity)
                         break;
                     case INpcEntity npc:
                         npc.Behavior.Update(npc);
+                        break;
+                    case IPlayerEntity player:
+
```
Eastrall: Oh, forgot to update the player's behavior.
TODO: add `player.Behavior.Update(player)`

PR #90, opened 2018-08-12
Chat commands/createitem

johmarjac:
This PR adds functionality to send World Messages to a user + I already edited the CreateItem Chat Command which required that packet.

Additionally I left a notice to consider loading the currently hardcoded text being sent to the player, loaded from a language dynamic file. So we might load a json file related to the world servers language setup and take the texts from there.

image

### DIFF
--- a/src/Rhisis.World/Game/Chat/CreateItemCommand.cs
+++ b/src/Rhisis.World/Game/Chat/CreateItemCommand.cs
@@ -1,4 +1,5 @@
 ﻿using NLog;
+using Rhisis.Core.Data;
 using Rhisis.Core.Structures.Game;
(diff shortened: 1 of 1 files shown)


johmarjac:
Okay, yeah, that makes much more sense. Thanks for the hint. I have replaced it with the defined text. Additionally this PR adds the SendWorldMsg packet then which is not being used then, I guess that is fine for future?

Eastrall:
All good. Let's merge this PR!

Skeatwin:
```SendWorldMsg``` in all case will be used one day or another. 
All not definded text will use this method. 
For exemple, if the administator want to send a message to all players ;)

PR #91, opened 2018-08-14
Messenger System

johmarjac:
Hi,

I´ve created this PR to discuss what I´ve done or should be doing.
As for now, I have created and added a new Component, the _MessengerComponent_, to the PlayerEntity, containing a _List<IPlayerEntity>_.

I´ve yet no idea how synchronizing with Database works as I have not yet looked into that yet. With that PR the MessengerSystem is added and the PacketType ADDFRIENDREQUEST is handled in that system. It checks the List<IPlayerEntity> if there is already the friend added, and if not it will send the Snapshot to the requested client. Nothing else is done yet and before I continue I wanted to discuss the code I´ve written so far here.


johmarjac:
No, I messed up commits, will fix this first.

PR #94, opened 2018-08-16
SendDefinedText proper Snapshot

johmarjac:
The game has two snapshots for DefinedText. The 0x0095 (which is the one used before this PR in Rhisis) is actually the one, that can pass additional text. It is used, when the text that should be printed is a format string that requires to be run through sprintf() or similar.

The other one, which is used for static text, is 0x0094 which really just sends the text id.

I have changed the current snapshottype 0x95 to 0x94. We need to have another SendDefinedText function which actually parses the formatted strings to be able to send dynamic text.


johmarjac:
@Eastrall The thing is, the few texts defined are in the wrong format. C# does not support string formats using %s or %d. I have found this https://stackoverflow.com/questions/14482341/c-net-and-sprintf-syntax/44390996#44390996 so far but have not yet tested it. It should work but I am not sure about special formats like %0.2f or some weird stuff like that. So I have that implemented on my local repository already but it needs testing first. Another alternative would be, replacing the formats in the defined texts to fit with C#'s string format syntax. I have counted them, its 264 lines only so it wouldnt be that bad to replace, but that however is not conform with "we stay at the original resource files".

Eastrall:
Interesting... But isn't the format made on the client side ? The servers sends the correct text id with the correct parameters and then the client formats it with a `sprintf` ?

johmarjac:
@Eastrall Unfortunately not only on the client-side as it seems. 
This is the CDPSrv::AddDefinedText() function from World-Server:
image

I am no C++ expert and I dont know what that _vsntprintf function does exactly but to me it looks like they put every argument based on the format string in parameters into one single string and on the client this string is parsed and "applied" to the final resource string.

Eastrall:
According to Microsoft documentation about `_vsnprintf` (https://msdn.microsoft.com/en-us/library/1kt27hek.aspx) I believe it behaves the same way as the `string.Format()` method in C#.

I don't know if it's possible, but instead of recoding the `_vsnprintf` function in C#, maybe we could simply replace all `%XX" by a `{i}` (where `i` is the index) and then apply a `string.Format` on the string.

johmarjac:
Yeah that is what the function is doing on StackOverFlow Link.
I just noticed you have your own dialog system with json files, that means we would need to convert all the dialogs to json anyways, why not converting the formatted dialogues to C# String.Format directly?!

johmarjac:
Most floats in the textClient.txt.txt are formatted like %.2f with a missing digit in front of the dot which results into not being recognized by that function of Stackoverflow. I am not familiar with Regular Expressions so someone else should probably better implement that. However the PR uses the proper Snapshot so that should be ready to merge if you want

chr1sto:
Just change the quantifier to * so it matches 0 or more.
%(\d*(\.\d+)?)?(d|f|s)

Eastrall:
Just tried the solution from @chr1sto on https://regex101.com/ It works like a charm! 😄

Thanks.

Eastrall:
Any news about this PR ?

johmarjac:
It ended up in me not understanding how to react to the above solution which I have told you some weeks a go, you said on Discord you would let me know how to do things so I can add them to the PR :) Also have I been very busy at work recently too unfortunately. I can barely follow the recent changes, hope to find some time in the next weeks.

Eastrall:
Yes true. I think we should create an extensions or a helper that formats the C style string into C# string 

⚠️ **not tested**
```cs
public string FormatCStyle(this string input, params object[] args)
{
    // based on : https://stackoverflow.com/questions/4098533/how-to-write-c-sharp-regular-expression-pattern-to-match-basic-printf-format-str
    string pattern = @"%(\d*(.\d+)?)?(d|f|s)";
    int count = 0;
    string result = Regex.Replace(input, pattern, m =>
    {
        var number = m.Groups["Number"].Value;
        var type = m.Groups["Type"].Value;
        return String.Concat("{", count++, "}");
    });

    return string.Format(result, args);
}
```

PR #95, opened 2018-08-16
Chat Command Authorization

johmarjac:
This PR adds authorization checks to the CreateItem, GetGold and Teleport Chat Commands where as the CreateItem and GetGold is only available as Administrator and Teleport only as GM or Admin.


johmarjac:
@Eastrall Yes I was also curious about using Tuple. I am fine by creating a small nested class. 👍

Eastrall:
All good! Merging...

PR #100, opened 2018-08-18
Administration tool

Eastrall:
This PR adds the Angular and ASP.NET Core project for the Rhisis Administration tool. It also changes the build chain, to build all the servers first and then build the adminsitration tool.


src/Rhisis.Admin/ClientApp/src/app/app.component.ts
```diff
@@ -3,8 +3,8 @@ import { Component } from '@angular/core';
 @Component({
   selector: 'app-root',
   templateUrl: './app.component.html',
-  styleUrls: ['./app.component.scss']
+  styleUrls: ['./app.component.css']
 })
 export class AppComponent {
-  title = 'RhisisAdmin';
+  title = 'app';
```
Skeatwin: Why ? This webapp is not Rhisis Web Admin ?

Eastrall: This component has been generated by the Visual Studio Angular template. For now there is only the basic client app solution. This component will be updated to include a better modular approach.
The title will be defined somewhere else.

src/Rhisis.Core/Helpers/RandomHelper.cs
```diff
@@ -58,4 +57,93 @@ public static int GenerateUniqueId()
             return ++_id;
         }
     }
+
```
Skeatwin: Can you please add a comment to describle the purpose of this class ? ^^
Eg: Helper given a quality random double by using Mersenne algorithm.

src/Rhisis.Core/Helpers/RandomHelper.cs
```diff
@@ -58,4 +57,93 @@ public static int GenerateUniqueId()
...
+    internal static class MersenneTwister
```
Skeatwin: In order to follow our class name convention, it would be preferable to name it ```MersenneTwisterHelper```.

src/Rhisis.Core/Helpers/RandomHelper.cs
```diff
@@ -58,4 +57,93 @@ public static int GenerateUniqueId()
...
+    {
+        private const int N = 624;
+        private const int M = 397;
+        private const uint MATRIX_A = 0x9908b0df;
+        private const uint UPPER_MASK = 0x80000000;
+        private const uint LOWER_MASK = 0x7fffffff;
+        private const uint TEMPERING_MASK_B = 0x9d2c5680;
+        private const uint TEMPERING_MASK_C = 0xefc60000;
```
Skeatwin: Upper const name are not compliant with our standard convention.
Please use pascal upper case. Eg: MatrixA, etc...

src/Rhisis.Core/Helpers/RandomHelper.cs
```diff
@@ -58,4 +57,93 @@ public static int GenerateUniqueId()
...
+            return (y << 7);
+        }
+        private static uint TEMPERING_SHIFT_T(uint y)
+        {
+            return (y << 15);
+        }
+        private static uint TEMPERING_SHIFT_L(uint y)
+        {
+            return (y >> 18);
+        }
```
Skeatwin: The same for function name, please use pascal upper case.

src/Rhisis.Core/Helpers/RandomHelper.cs
Eastrall: The `MerseenTwister` is a pseudo-random number generator used for generating random number in the Rhisis solution. This class has been taken from another project, I've lost the credit, but I have to find it and place it again.
The `MerseenTwister` class shouldn't be used outside the `Rhisis.Core` assembly, this is why he has the `internal` access modifier.
As for the name, since it is an external code, the coding principles doesn't really apply to it. It's not our code.

src/Rhisis.Core/Helpers/RandomHelper.cs
Eastrall: This class is just a random helper that provides useful methods to generate random number using the MerseenTwister algorithm. I don't know if it's really important to specify which algorithm we use in the docs, since it is "just" a wrapper.

src/Rhisis.Core/Helpers/RandomHelper.cs
Skeatwin: OK but only because it's your birthday... 😄

src/Rhisis.Core/Helpers/RandomHelper.cs
Skeatwin: Ok, fews words just to helpful readers.

Skeatwin:
As ```MerseenTwister``` class is from external, I think we can put all external classes in an ```External``` folder/namespace.
This will allow us to keep original files without changes them. 
Please be carefull to sync class name and file name (which is not correct for now).

PR #101, opened 2018-08-22
Improve DependencyContainer and add unit tests

Eastrall:
This PR improves the `DependencyContainer` and adds some unit tests for it.

### DIFF
--- a/src/Rhisis.Core/DependencyInjection/DependencyContainer.cs
+++ b/src/Rhisis.Core/DependencyInjection/DependencyContainer.cs
@@ -6,11 +6,16 @@ using System.Linq;
 
 namespace Rhisis.Core.DependencyInjection
 {
-    public class DependencyContainer : Singleton<DependencyContainer>
+    public sealed class DependencyContainer : Singleton<DependencyContainer>, IDisposable
     {
         private IServiceCollection _services;
-        private IServiceProvider _serviceProvider;
+        private ServiceProvider _serviceProvider;
         private bool _isInitialized;
+
+        /// <summary>
+        /// Gets the number of registered services.
+        /// </summary>
+        public int Count => this._services.Count;
 
         /// <summary>
         /// Creates a new <see cref="DependencyContainer"/> instance.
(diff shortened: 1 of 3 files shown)


src/Rhisis.Core/DependencyInjection/DependencyContainer.cs
```diff
@@ -29,21 +34,30 @@ public DependencyContainer Initialize()
             if (this._isInitialized)
                 throw new InvalidOperationException("Dependency container is already initialized.");
 
+            if (this._services == null)
+                throw new InvalidOperationException("Service collection is not initialized.");
```
Skeatwin: This case will never hapen as it's not a static method and the constructor intance always this variable.

src/Rhisis.Core/DependencyInjection/DependencyContainer.cs
```diff
@@ -57,10 +71,7 @@ public DependencyContainer Initialize(IServiceCollection serviceCollection)
         /// <returns></returns>
         public IServiceProvider BuildServiceProvider()
         {
-            if (!this._isInitialized)
-                throw new InvalidOperationException("Dependency container is not initialized. Please call the Initialize() method first.");
-
-            if (this._serviceProvider == null)
+            if (this._serviceProvider == null && this._services != null)
```
Skeatwin: Same here for ```this._services``` ^^

src/Rhisis.Core/DependencyInjection/DependencyContainer.cs
```diff
@@ -71,6 +82,47 @@ public IServiceProvider BuildServiceProvider()
...
+        /// <summary>
+        /// Register a new service.
+        /// </summary>
+        /// <param name="implementationType">Service implementation type</param>
+        /// <param name="serviceType">Service type</param>
+        /// <param name="serviceLifetime">Service life time</param>
+        public void Register(Type implementationType, Type serviceType, ServiceLifetime serviceLifetime = ServiceLifetime.Transient)
+        {
+            if (this._services == null)
+                throw new InvalidOperationException("Cannot register dependency. ServiceCollection has not been initialized. Please call the Initialize() method.");
```
Skeatwin: Again this cas will never happen, but may be I've miss something ^^'

src/Rhisis.Core/DependencyInjection/DependencyContainer.cs
Skeatwin: Correct, I've just see it now ! I was sure to forget something 😕

src/Rhisis.Core/DependencyInjection/DependencyContainer.cs
Skeatwin: My bad, this case can be null 👍

PR #115, opened 2018-10-02
Remove packets definition from Rhisis.Network

Eastrall:
This PR removes the packet definition classes from the `Rhisis.Network` assembly and reads directly the packet data inside the handlers.


Eastrall:
There is some unused variables as the codacy analyse report says. In my opinion, we can ignore them because it is mostly variables related to packet structure. (unused for now)

PR #123, opened 2018-10-27
Update game system management.

Eastrall:
This PR updates and improves the game system management. In the previous version, to create a system, you would need to create a new class and inherit from `SystemBase` or `NotifiableSystemBase` wether you want to create an updatable system or be notifiable one and then implement the `Execute` method.
To avoid unnecessary complexity, we removed all of this classes and only keep the `ISystem` interface. Plus, we've updated the `[SystemAttribute]` so you can specify the type of the system.

There is a exemple of how to create a system now:
```cs
[System(SystemType.Notifiable)]
public class MySystem : ISystem
{
     // Specify the type of entities that can execute the system
     public WorldEntityType Type => WorldEntityType.Player;

     // Put your game logic here
     public void Execute(IEntity entity, SystemEventArgs e)
     {
     }
}
```
By default, a `System` is updatable, you need to specify the `Notifiable` to create a notifiable system.

Finally, last improvement to this system. In previous version, we would update the systems manually (for updatable systems). This method was looping through the `Context.Systems` array and had a O(N) complexity. And this loop was also inside the update entities loop. So, for X entities and N systems, we would have a complexity of O(X * N), which is very bad for the CPU.
Now, during the initialization process of the WorldServer, the `SystemManager` class will create one and only one method that will call the updatable systems one by one without any loop. This will result in a complexity of O(1) which is really good.


Eastrall:
Thanks @Steve-Nzr! Besides performance improvements, I hope it will make system creation much easier than it was before ! 😄

PR #138, opened 2018-12-03
Implemented basic WASD movement

Kaev:
Implemented basic WASD movement.
Motion stuff still needs to be implemented before we can add it here.
I'm not sure if we have to/should implement the delay stuff.
Death and flying stuff also needs to be implemented before this can be finished.

### DIFF
--- a/src/Rhisis.Network/Packets/World/OnPlayerMovedPacket.cs
+++ b/src/Rhisis.Network/Packets/World/OnPlayerMovedPacket.cs
@@ -1,0 +1,90 @@
+﻿using System;
+using Ether.Network.Packets;
+using Rhisis.Core.Structures;
+
+namespace Rhisis.Network.Packets.World
+{
+    /// <summary>
+    /// Defines the <see cref="OnPlayerMovedPacket"/> structure.
+    /// </summary>
+    public struct OnPlayerMovedPacket : IEquatable<OnPlayerMovedPacket>
+    {
+        /// <summary>
+        /// Gets the BeginPosition.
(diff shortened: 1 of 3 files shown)


src/Rhisis.Network/Packets/World/OnPlayerMovedPacket.cs
```diff
@@ -0,0 +1,90 @@
...
+
+        /// <summary>
+        /// Gets the DestinationPosition.
+        /// </summary>
+        public Vector3 DestinationPosition { get; }
+
+        /// <summary>
+        /// Gets the SpeedFactor.
+        /// </summary>
+        public float SpeedFactor { get; }
```
Kaev: This might be angle instead of SpeedFactor.

Eastrall: Yes indeed. If I understand correctlyt the `OnPlayerMoved2()` (method that handles the fly moves) from the [official files](https://github.com/domz1/SourceFlyFF/blob/ce4897376fb9949fea768165c898c3e17c84607c/Program/WORLDSERVER/DPSrvr.cpp#L2467), the `f` field might be the angle since we also have the `fAngleX` right after it. 
Not entierly sure about that though...

Kaev: I'll fix that as soon as i'm home.

Eastrall:
Closing PR. Solved in PR #204

PR #139, opened 2018-12-05
Mail system

Kaev:
This PR adds the mail system.
This also adds the text packets (#140, #94) because i needed them for sending a proper error message to the client (like official files).
It also contains a lot of changes in different areas of Rhisis that i noticed while working on the mailbox stuff.


src/Rhisis.World/Systems/Mailbox/MailboxSystem.cs
```diff
@@ -0,0 +1,161 @@
...
+
+            var receiverId = 0;
+            Game.Structures.Item item = null;
+            var itemQuantity = 0;
+            var neededGold = 500; // should be a config value
+
+            // Receiver is offline
+            if (receiverEntity is null)
+            {
+                var database = DependencyContainer.Instance.Resolve<IDatabase>();
```
Eastrall: You created a new connection to the database, but you don't close it. You can use the `database.Dispose()` method to close the database connection, or use an `using` statement:
```csharp
using (var database = DependencyContainer.Instance.Resolve<IDatabase>())
{
     // Do stuff
} // Call dispose automatically
```

src/Rhisis.World/Systems/Mailbox/MailboxSystem.cs
```diff
@@ -0,0 +1,161 @@
...
+            if (e.ItemSlot < InventorySystem.InventorySize)
+            {
+                item = player.Inventory.Items[e.ItemSlot];
+                if (e.ItemQuantity > item.Quantity)
+                    return; // Is there an error packet?
+
+                itemQuantity = item.Quantity;
+            }
+            else // Itemslot is not a valid inventory slot
+            {
```
Eastrall: You can log an error here and cancel the operation with a `return`

src/Rhisis.World/WorldServer.cs
```diff
@@ -93,5 +93,16 @@ public IPlayerEntity GetPlayerEntity(int id)
...
         }
+
+        /// <summary>
+        /// Gets a player entity by his name
+        /// </summary>
+        /// <param name="name"></param>
+        /// <returns></returns>
+        public IPlayerEntity GetPlayerEntity(string name)
+        {
+            WorldClient client = this.Clients.FirstOrDefault(x => x.Player.Object.Name == name);
```
Eastrall: To compare two strings, I suggest you to use the:
```csharp
x => x.Player.Object.Name.Equals(name, StringComparison.OrdinalIgnoreCase)
```

src/Rhisis.World/Systems/Mailbox/MailboxSystem.cs
```diff
@@ -0,0 +1,161 @@
...
+using Rhisis.World.Packets;
+using Rhisis.World.Systems.Inventory;
+using Rhisis.World.Systems.Mailbox.EventArgs;
+
+namespace Rhisis.World.Systems.Mailbox
+{
+    [System(SystemType.Notifiable)]
+    public class MailboxSystem : ISystem
+    {
+        private static readonly ILogger Logger = LogManager.GetCurrentClassLogger();
```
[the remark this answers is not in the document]
Kaev: Then i have to add a reference to Microsoft.Extensions.Logging - I changed it anyway.

src/Rhisis.Database/Entities/DbMail.cs
```diff
@@ -0,0 +1,53 @@
...
+    public sealed class DbMail : DbEntity
+    {
+        public DbCharacter Sender { get; set; }
+        public DbCharacter Receiver { get; set; }
+        public int Gold { get; set; }
+        public DbItem Item { get; set; }
+        public short ItemQuantity { get; set; }
+        public string Title { get; set; }
+        public string Text { get; set; }
+        public bool Read { get; set; }
```
Eastrall: I was thinking that maybe this property could be more a `IsOpened` boolean, right ?

src/Rhisis.World/Systems/Mailbox/MailboxSystem.cs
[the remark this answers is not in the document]
Eastrall: It's true that other systems still have the NLog `LogManager` reference and this is something we want to change in the future. In my opinion, it's better to use the abstraction rather than the concret reference.

src/Rhisis.World/Packets/TextMessagePackets.cs
```diff
@@ -1,48 +1,65 @@
...
         {
             using (var packet = new FFPacket())
             {
-                // ifdef __S_SERVER_UNIFY
-                packet.StartNewMergedPacket(0, SnapshotType.TEXT);
-                packet.Write((byte)0x02); //const BYTE TEXT_DIAG = 0x02 https://github.com/domz1/SourceFlyFF/blob/ce4897376fb9949fea768165c898c3e17c84607c/Program/_Network/MsgHdr.h#L1583
-                // else __S_SERVER_UNIFY
-                packet.StartNewMergedPacket(0, SnapshotType.DIAG_TEXT);
-                // endif __S_SERVER_UNIFY
+                if (MailboxSystem.TEXT_TYPE == TextType.TEXT_DIAG)
```
Eastrall: This condition will always be true since `MailboxSystem.TEXT_TYPE = TextType.TEXT_DIAG`. So, snapshot `SnapshotType.DIAG_TEXT` will never be called. Is it the intended behavior ?

src/Rhisis.World/Systems/Mailbox/MailboxSystem.cs
```diff
@@ -17,6 +18,9 @@ public class MailboxSystem : ISystem
     {
         private static readonly ILogger Logger = DependencyContainer.Instance.Resolve<ILogger<MailboxSystem>>();
 
+        public static readonly int MAX_MAIL = 50;
+        public static readonly TextType TEXT_TYPE = TextType.TEXT_DIAG;
```
Eastrall: I suggest you to use PascalCase for constants as explained here in the [C# naming convention](https://blogs.msdn.microsoft.com/mschray/2013/05/06/practical-naming-standards-for-c/). The only exception made for Rhisis, is the `PacketType` enum that has been converted from the official files.

src/Rhisis.World/Packets/TextMessagePackets.cs
Kaev: Yes, currently it's intended. The official server files have this aswell (the __S_SERVER_UNIFY define) and i'm not sure yet if this is deactivated in some clients.

src/Rhisis.World/Systems/Mailbox/MailboxSystem.cs
Kaev: Will change that, thought you might want all enums uppercase that we take out of the official files.

Eastrall: Not entierly sure about enum names, since what we actually convert is a set of defines. This is why I just keep the enum's content has it is in the official files.

bin/config/world.json
```diff
@@ -12,6 +12,7 @@
     "host": "127.0.0.1",
     "port": 15000
   },
+  "mailshippingcost": 500,
```
Eastrall: JSON property names should be in `camelCase`; In other words, this property should be: `mailShippingCost`.

I like the idea of making the shipping cost configurable!

src/Rhisis.World/Systems/Mailbox/MailboxSystem.cs
```diff
@@ -66,91 +67,111 @@ private void GetMails(IPlayerEntity player, QueryMailboxEventArgs e)
...
+            using (var database = DependencyContainer.Instance.Resolve<IDatabase>())
+            {
+                receiver = database.Characters.Get(x => x.Name == e.Receiver);
+                if (receiver is null)
+                {
+                    WorldPacketFactory.SendAddDiagText(player, "TID_MAIL_UNKNOW"); // Get text of TID_MAIL_UNKNOW here
+                    return;
+                }
+                sender = database.Characters.Get(x => x.Id == player.PlayerData.Id);
+                receiverMailQuantity = database.Mails.Count(x => x.Receiver == receiver);
```
Eastrall: Maybe you could `Include()` the `Mail` property in the `CharacterRepository` ? Instead of making a query through all mails in database, you could just make: `receiver.Mails.Count()`.

src/Rhisis.World/Systems/Mailbox/MailboxSystem.cs
```diff
@@ -66,91 +67,111 @@ private void GetMails(IPlayerEntity player, QueryMailboxEventArgs e)
...
-            else // Itemslot is not a valid inventory slot
-            {
-                Logger.LogError($"ItemSlot {e.ItemSlot} is not a valid inventory slot.");
-                return; // Is there an error packet?
-            }
-            
+                // Calculate gold amount
+                if (e.Gold < 0)
+                {
+                    WorldPacketFactory.SendAddDiagText(player, "TID_GAME_LACKMONEY"); // Get text of TID_GAME_LACKMONEY here
```
Eastrall: Is there any `DefineText.TID_GAME_LACKMONEY` enum field ?

Kaev: The ID isn't used here, we actually need to send the text of TID_GAME_LACKMONEY.

Eastrall: Oh wow, really ? 😨

src/Rhisis.World/Systems/Mailbox/MailboxSystem.cs
```diff
@@ -0,0 +1,288 @@
...
+                var receiver = database.Characters.Get(x => x.Id == player.PlayerData.Id);
+                var sender = database.Characters.Get(x => x.Id == 2);
+                if (receiver != null)
+                    WorldPacketFactory.SendMailbox(player, receiver.ReceivedMails.Where(x => !x.IsDeleted).ToList());
+            }
+        }
+
+        private void SendMail(IPlayerEntity player, QueryPostMailEventArgs e)
+        {
+            // TODO: If mailbox is too far away: return;
```
Eastrall: Can you send a mail even if you are far away from the mail box ? Does the client checks that ?

Kaev: Never trust the client. You can just send the send packet and it would allow you to send mail if this isn't checked.

johmarjac: We need a NpcHelper class or something which has a function like IsNearNpc

Eastrall: Do we have the object id of the mail box npc ?

johmarjac: The npc must be in the entity list if the player is near, so we can simply check the npc id in the defines whether it is in entity list and whether it is then near enough.

src/Rhisis.World/Systems/PlayerData/PlayerDataSystem.cs
```diff
@@ -0,0 +1,69 @@
...
+                case QueryPlayerData2EventArgs queryPlayerData2Event:
+                    GetPlayerData(playerEntity, queryPlayerData2Event);
+                    break;
+            }
+        }
+
+        private void GetPlayerData(IPlayerEntity player, QueryPlayerDataEventArgs e, bool send = true)
+        {
+            var worldServer = DependencyContainer.Instance.Resolve<IWorldServer>();
+            var playerEntity = worldServer.GetPlayerEntity(e.PlayerId);
```
Eastrall: I think this will always return `null`. The `GetPlayerEntity(int id)` method, returns a entity (world) by his id. Not the "real" player id (database).
These are two different things:
- `player.Id`: Gives you the **entity id** (object id)
- `player.PlayerData.Id`: Gives you the player's id (stored in database)

src/Rhisis.World/WorldServer.cs
```diff
@@ -98,5 +98,11 @@ public IPlayerEntity GetPlayerEntity(string name)
             WorldClient client = this.Clients.FirstOrDefault(x => x.Player.Object.Name.Equals(name, StringComparison.OrdinalIgnoreCase));
             return client?.Player;
         }
+
+        public IPlayerEntity GetPlayerEntityByCharacterId(uint id)
```
Eastrall: 👍

src/Rhisis.Network/Packets/ModeType.cs
```diff
@@ -0,0 +1,39 @@
+﻿using System;
+
+namespace Rhisis.Network.Packets
+{
+    [Flags]
+    public enum ModeType : uint
```
Eastrall: You should add this flags to the `Rhisis.Core` assembly

Eastrall:
@Kaev you have some conflicts with this files:

- `src/Rhisis.World/Packets/BattlePackets.cs`
- `src/Rhisis.World/Packets/WorldPacketFactory.cs`

Can you fix this conflict and merge your commits ? thanks 😄

src/Rhisis.Database/Repositories/Implementation/CharacterRepository.cs
```diff
@@ -23,7 +23,21 @@ protected override IQueryable<DbCharacter> GetQueryable(DbContext context)
         {
             return base.GetQueryable(context)
                 .Include(x => x.User)
-                .Include(x => x.Items);
+                .Include(x => x.Items)
+                .Include(x => x.Items)
+                    .ThenInclude(x => x.Character)
```
Eastrall: Do we actually need to include the owner of the item ? Are we going to do something like: `character.Items[0].Character` ?

src/Rhisis.World/Packets/TextMessagePackets.cs
```diff
@@ -0,0 +1,67 @@
...
+                player.Connection.Send(packet);
+            }
+        }
+
+        /// <summary>
+        /// Shows a defined text at the client.
+        /// </summary>
+        /// <param name="player"></param>
+        /// <param name="textId"></param>
+        public static void SendAddDefinedText(IPlayerEntity player, DefineText textId)
```
Eastrall: Why not `SendDefinedText` ? Might be more readable than `SendAddDefinedText` no ?

Kaev: Took the name from the official functions, we can name it SendDefinedText though

src/Rhisis.Database/Repositories/Implementation/CharacterRepository.cs
Kaev: Probably not, will remove that

src/Rhisis.World/Systems/Mailbox/MailboxSystem.cs
```diff
@@ -0,0 +1,324 @@
...
+                        return;
+                    }
+
+                    if (inventoryItem.ExtraUsed != 0)
+                    {
+                        WorldPacketFactory.SendAddDiagText(player, textClient["TID_GAME_CANNOT_POST"]);
+                        return;
+                    }
+
+                    if (inventoryItem.Data.ItemKind3 == ItemKind3.QUEST)
```
Eastrall: I think you should merge the statements together to prevent code duplication:
```csharp
if (inventoryItem.IsEquipped() || inventoryItem.ExtraUsed != 0 || inventoryItem.Data.ItemKind3 == ItemKind3.QUEST)
{
     WorldPacketFactory.SendAddDiagText(player, textClient["TID_GAME_CANNOT_POST"]);
     return;
}
```

src/Rhisis.World/Systems/Mailbox/MailboxSystem.cs
```diff
@@ -0,0 +1,324 @@
...
+
+                checked
+                {
+                    try
+                    {
+                        player.PlayerData.Gold += (int)mail.Gold;
+                        mail.HasReceivedGold = true;
+                    }
+                    catch (OverflowException)
+                    {
```
Eastrall: Log an error here, just in case... 😄

Eastrall:
Looking quite good to me. I guess we can merge this system once you have done the last changes! 😄

Kaev:
Done :)

PR #142, opened 2018-12-07
Add SendAddDamage Packet

johmarjac:
This PR (will)contain(s) an implementation of the SNAPSHOTTYPE_DAMAGE Packet. The first commit shows a prototype implementation, however this one is not tested. Further things should be discussed here e.g. where to put the Enum AttackFlags etc.

### DIFF
--- a/src/Rhisis.World/Game/Common/AttackFlags.cs
+++ b/src/Rhisis.World/Game/Common/AttackFlags.cs
@@ -1,0 +1,27 @@
+﻿using System;
+
+namespace Rhisis.World.Game.Common
+{
+    [Flags]
+    public enum AttackFlags : uint
+    {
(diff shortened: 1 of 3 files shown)


Eastrall:
Maybe we could create a new folder named `Flags` or `Common` inside the `Rhisis.World/Game/` folder ? This folder could also contain the `MovingFlags` and other common game stuff.

Eastrall:
Can you add this statement to the the `ProcessMeleeAttack()` method (https://github.com/Eastrall/Rhisis/blob/develop/src/Rhisis.World/Systems/Battle/BattleSystem.cs#L51) ? As for the `AttackFlag` I think you can use the `AF_GENERIC`.

The packet will be already in place and we can test it easily. 😄

PR #143, opened 2018-12-07
Added SetTargetHandler

johmarjac:
This PR implements the SetTargetHandler which is a packet being sent from both Server and Client when the player targets an entity in the world (NPC's, Monsters, Objects etc.)

Its not fully tested, but initially worked for NPC's and Monsters. 
I´ve setup a InteractionSystem to process the packet and a InteractionComponent to store the targeted entity. 
There is also support for deselecting the target (i.e. hitting Escape when a monster was selected which results into clearing the target selection).

I thought it was good idea to implement now because I want to create ChatCommands when working with the BattleSystem. So we can use the currently targeted entity in the chat command to do stuff with.

However, let's discuss about this implementation :)


src/Rhisis.World/Systems/Interaction/InteractionSystem.cs
```diff
@@ -0,0 +1,66 @@
...
+using Rhisis.World.Game.Core.Systems;
+using Rhisis.World.Game.Entities;
+using Rhisis.World.Systems.Interaction.EventArgs;
+
+namespace Rhisis.World.Systems.Interaction
+{
+    [System(SystemType.Notifiable)]
+    public class InteractionSystem : ISystem
+    {
+        private static readonly ILogger Logger = LogManager.GetCurrentClassLogger();
```
Eastrall: Please use the `DependencyContainer.Instance.Resolve<ILogger<InteractionSystem>>();` instead of the `GetCurrentClassLogger()` from NLog.

src/Rhisis.World/Systems/Interaction/InteractionSystem.cs
```diff
@@ -0,0 +1,66 @@
...
+        /// </summary>
+        /// <param name="player"></param>
+        /// <param name="e"></param>
+        private void SetTarget(IPlayerEntity player, SetTargetEventArgs e)
+        {
+            var targetEntity = player.FindEntity<IEntity>(e.TargetId);
+            if (targetEntity == null)
+                return;
+
+            if (e.Clear == 2)
```
Eastrall: What does the `2` mean ? Is there a constant somewhere describing this ?

johmarjac: 2 means select,
1 deselect.
Yeah I see your issue, it's not describing well what it is for. Lemme see if I can do an Enum for that, what do you think?

Kaev: An enum or two constants - Everything is better than a magic number.

Eastrall: I agree with @Kaev. Maybe two constant values should be enough.

johmarjac: I´ve introduced an enumeration for that.

Eastrall:
All good to me. Let's merge.

PR #154, opened 2018-12-13
TaskbarSystem (Shortcuts, Actionslot)

johmarjac:
This PR adds the Taskbar System.
As of now (and what I am aware of) : The Taskbar System handles Shortcuts (i.e. Chat Scripts, Items, Skills) that are being dragged onto the Taskbar of the player. 

As of writing this, the left side of the taskbar (which is called the Applet Bar) and the middle (which is called the Item Bar) are implemented for communication with the game client. Adding Shortcuts and Removing Shortcuts are working. However, they are not yet saved to the database which is the next step as for the right side (called the Actionslot) I first need to be able to put skills in there to (not implement) but test it. (It's not yet implemented).

The Taskbar System is not yet finished so I dont want it to be merged as it would not make sense anyways at this state.

However feedback n' stuff is greatly appreciated and requested.

Oh and btw: Documentation will be added as the last commit if the system is finished.

### DIFF
--- a/src/Rhisis.Core/Common/Game/Structures/Shortcut.cs
+++ b/src/Rhisis.Core/Common/Game/Structures/Shortcut.cs
@@ -1,7 +1,6 @@
 ﻿using Ether.Network.Packets;
-using Rhisis.Core.Common;
 
-namespace Rhisis.World.Game.Structures
+namespace Rhisis.Core.Common.Game.Structures
 {
     public class Shortcut
     {
--- a/src/Rhisis.Core/Common/Game/Structures/Shortcut.cs
+++ b/src/Rhisis.Core/Common/Game/Structures/Shortcut.cs
@@ -6,11 +6,11 @@ namespace Rhisis.Core.Common.Game.Structures
     {
         public int SlotIndex { get; }
 
-        public ShortcutType ShortcutType { get; }
+        public ShortcutType Type { get; }
 
         public uint ObjId { get; }
 
-        public ShortcutObjType ObjType { get; }
+        public ShortcutObjectType ObjectType { get; }
 
         public uint ObjIndex { get; }
 
@@ -20,12 +20,12 @@ namespace Rhisis.Core.Common.Game.Structures
 
         public string Text { get; }
 
-        public Shortcut(int slotIndex, ShortcutType shortcutType, uint objId, ShortcutObjType shortcutObjType, uint objIndex, uint userId, uint objData, string text)
+        public Shortcut(int slotIndex, ShortcutType type, uint objId, ShortcutObjectType shortcutObjectType, uint objIndex, uint userId, uint objData, string text)
         {
             SlotIndex = slotIndex;
-            ShortcutType = shortcutType;
+            Type = type;
             ObjId = objId;
-            ObjType = shortcutObjType;
+            ObjectType = shortcutObjectType;
             ObjIndex = objIndex;
             UserId = userId;
             ObjData = objData;
@@ -35,14 +35,14 @@ namespace Rhisis.Core.Common.Game.Structures
         public void Serialize(INetPacketStream packet)
         {
             packet.Write(SlotIndex);
-            packet.Write((uint)ShortcutType);
+            packet.Write((uint)Type);
             packet.Write(ObjId);
-            packet.Write((uint)ObjType);
+            packet.Write((uint)ObjectType);
             packet.Write(ObjIndex);
             packet.Write(UserId);
             packet.Write(ObjData);
 
-            if (ShortcutType == ShortcutType.Chat)
+            if (Type == ShortcutType.Chat)
                 packet.Write(Text);
         }
     }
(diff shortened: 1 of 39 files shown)


src/Rhisis.Core/Common/ShortcutType.cs
```diff
@@ -0,0 +1,26 @@
...
+        Item,
+        Skill,
+        Object,
+        Chat,
+        Skillfun,
+        Emoticon,
+        Lordskill
+    }
+
+    public enum ShortcutObjType : uint
```
Eastrall: Can you rename it in `ShortcutObjectType` ? This is more readable than `ShortcutObjType` in my opinion. 😄

src/Rhisis.World/Game/Components/TaskbarAppletContainerComponent.cs
```diff
@@ -0,0 +1,69 @@
...
+        {
+            if (slotIndex < 0 || slotIndex >= TaskbarSystem.MaxTaskbarApplets)
+                return;
+
+            Shortcuts[slotIndex] = null;
+        }
+
+        public int Count
+        {
+            get
```
Eastrall: Just a little thing here, try using C# 7 getters for code consistency.😉
```csharp
public int Count => this.Shortcuts.Count(x => x != null && x.ShortcutType != ShortcutType.None);
```

src/Rhisis.Core/Common/Game/Structures/Shortcut.cs
```diff
@@ -0,0 +1,49 @@
+﻿using Ether.Network.Packets;
+
+namespace Rhisis.Core.Common.Game.Structures
+{
+    public class Shortcut
+    {
+        public int SlotIndex { get; }
+
+        public ShortcutType ShortcutType { get; }
```
Eastrall: Maybe you can rename this property into `Type` only. Because when you are going to use a `Shortcut` object/instance, and you want to access this property you will be writing:
```csharp
var shortcut = new Shortcut(...);

if (shortcut.ShortcutType == ....)
    ...
```
When you read this, you are writing "shortcut" twice. But in fact, you are already in the `shortcut` context, so you could just write:

```csharp
var shortcut = new Shortcut(...);

if (shortcut.Type == ....)
    ...
```
Here you simply access the shortcut type. It's more "natural" when you read it. 😄 
Be careful when you choose your variable names. Always try to think about how you are going to write the code to access this property/method.

src/Rhisis.Core/Common/ShortcutType.cs
```diff
@@ -0,0 +1,26 @@
...
+        None,
+        Etc,
+        Applet,
+        Motion,
+        Script,
+        Item,
+        Skill,
+        Object,
+        Chat,
+        Skillfun,
```
Eastrall: This enum should be in `PascalCase` : `SkillFun`

Eastrall:
That's a good begining. I can see some code duplication between the `AddTaskbarAppletPacket` and the `AddTaskbarItemPacket`. Maybe a `AddTaskbarItemPacket` could inherit from `AddTaskbarAppletPacket` ?

src/Rhisis.Core/Common/ShortcutType.cs
johmarjac: Yup, will do :)

johmarjac:
@Eastrall Will do! :)

src/Rhisis.Core/Common/Game/Structures/Shortcut.cs
johmarjac: Great, yeah you´re right, that's much better reading 👍 
Thanks

src/Rhisis.World/Packets/SpawnPackets.cs
```diff
@@ -206,6 +207,15 @@ public static void SendPlayerSpawn(IPlayerEntity player)
...
                 packet.Write(0); // count
+                
+                player.Connection.Send(packet);
+            }
+                                   
+            // Taskbar
+            using (var packet = new FFPacket())
+            {
+                packet.StartNewMergedPacket(player.Id, SnapshotType.TASKBAR);
+                player.Taskbar.Serialize(packet);
```
Eastrall: Can't you just start a new merged packet inside the previous one instead of creating a new instance of a packet ?
Like
```csharp
// buffs
packet.Write(0); //count

// Taskbar
packet.StartnewMergedPacket(player.Id, SnapshotType.TASKBAR);
player.Taskbar.Serialize(packet);

player.Connection.Send(packet);
```

johmarjac: image
You mean like so? I actually tried that before I did what I did now. I just tested it right now again, it crashs the client then.

johmarjac: Okii Dokee

Eastrall:
@johmarjac Can you fix the conflicts so we can merge your PR to test in the test server ? 😄

Issue #160, opened 2019-01-02
Drop system


Eastrall:
## Description

The drop system allow the player/monsters to place(drop) items on the map. A player can simply drag'n'drop an item from the inventory or a bag to the map to drop it and a monster drops items on the map when they die via the Battle System or a command.
There are three types of drops:

- Permenant drops (items dropped by a player)
- Temporary drops (items dropped by a monster when they die): These drops are linked to the player who killed the monster for some time. (time to define)
- Respawn drops (quest items)

**Note:** Gold is also considered as a dropped item.

## Roadmap

- [x] Create the `ItemEntity` that will reprensent an dropped item on the map
- [x] Create the `DropSystem` class
- [x] Create a permanent drop
- [x] Temporary drop
- [x] Respawn drop
- [x] Pick-up a drop (item or gold) (https://github.com/Eastrall/Rhisis/commit/f31862ecee70bb21cf062386e9c30e958aa2afac, https://github.com/Eastrall/Rhisis/commit/3f4eba25782a15b1fa372313fdac470e42314f16)
- [x] Player drop item from inventory (https://github.com/Eastrall/Rhisis/commit/fcb0fe617335a38bd227f49a1d519df8747551bb)

Issue #29, opened 2018-02-21
Battle System


Eastrall:
### Description

The battle system is one of the main system of the FLYFF game. In this system you can cause damages to the monsters and other players (PK, arena and PVP).

#### Base

- [x] Receive/handle melee attack packet
- [x] Make the monster follow the player
- [x] Make the monster unfollow the player if he runs away (out of monster circle radius; maybe 20f)
- [x] Calculate melee damage (With weapons, bonuses and defense)
- [x] Decrease HP from monster/player
- [x] Add monster AI for battle
- [x] Load character experience table (7abd073)
- [x] Give exp to player (450d864)
- [x] Level up (450d864)
- [x] Give attribute points on level up (450d864)
- [x] Give skill points on level up (450d864)

PR #162, opened 2019-01-06
Drop System

Eastrall:
This PR includes the `DropSystem` and monster gold item drops.  It covers issue #29 and #160.

The amount of gold and items are loaded from the `propMoverEx.inc` file. The following statements are loaded:
- `DropGold()`
- `DropItem()`
- `DropKind()`
- `Maxitem`

The `DropSystem` allows an entity to drop an item or gold to the current layer.

The system has been design to handle any type of scenario. 

You can drop an item or gold simply by calling the `DropSystem` instance: `entity.NotifySystem<DropSystem>(...)`. The arguments are the following:

- `DropItemEventArgs`:
  - `ItemBase` : Item to drop
  - `Owner` : Entity owner
- `DropGoldEventArgs`
  - `GoldAmount`: Gold amount to drop
  - `Owner`: Entity owner


bin/config/world.json
```diff
@@ -7,6 +7,11 @@
     "WI_WORLD_MADRIGAL"
   ],
   "language": "en",
+  "rates": {
+    "gold": 10,
+    "experience": 1,
+	"drop": 10
```
Kaev: shouldn't it be 1? for gold aswell

Eastrall: Oops, that was a test value... I'll revert it.

Issue #161, opened 2019-01-02
Customize system


Eastrall:
## Description

FLYFF has a customization system that allows the player to change his hair style and select a new face. The system is triggered by NPC at Northern Flarine.
![image](https://user-images.githubusercontent.com/4021025/50601093-77010080-0eb3-11e9-9fee-b47c5a79d4e5.png)

More informations about the system here: http://flyff-wiki.webzen.com/wiki/Customizing

## Roadmap

- [x] Change hair style and color
- [x] Change face style
- [x] Remove gold from player
- [x] Use coupon

Eastrall:
Implemented in PR #166 and #168

PR #166, opened 2019-01-12
Facial and Hair Customization with Gold

johmarjac:
This PR adds (according to #161 ) the customization system or at least a good part of it.

Due to problems with the ItemContainer for the players inventory I was not yet able to implement the customization via coupon items. We gotta refactor the inventory item container before this can be added properly.

However changing face id, hair id and color works fine. Feedback please.

### DIFF
--- a/src/Rhisis.Core/Structures/Configuration/WorldConfiguration.cs
+++ b/src/Rhisis.Core/Structures/Configuration/WorldConfiguration.cs
@@ -74,7 +74,7 @@ namespace Rhisis.Core.Structures.Configuration
         public uint MailShippingCost { get; set; }
 
         [DataMember(Name = "customization")]
-        public StyleCustomization Customization { get; set; }
+        public StyleCustomization Customization { get; set; } = new StyleCustomization();
 
         /// <summary>
         /// Creates a new <see cref="WorldConfiguration"/> instance.
@@ -86,7 +86,6 @@ namespace Rhisis.Core.Structures.Configuration
             this.ISC = new ISCConfiguration();
             this.Rates = new WorldRates();
             this.MailShippingCost = DefaultMailShippingCost;
-            this.Customization = new StyleCustomization();
         }
     }
 }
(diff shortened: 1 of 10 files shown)


src/Rhisis.Core/Structures/Configuration/WorldConfiguration.cs
```diff
@@ -66,9 +69,30 @@ public class WorldConfiguration : BaseConfiguration
...
+        /// Gets or sets mail shipping costs.
+        /// </summary>
         [DataMember(Name = "mailShippingCost")]
         public uint MailShippingCost { get; set; }
 
+        /// <summary>
+        /// Gets or sets the costs for changing characters face.
+        /// </summary>
+        [DataMember(Name = "changeFaceCost")]
+        public uint ChangeFaceCost { get; set; }
```
Eastrall: it woud not be better to have an object like : `StyleCustomization` with this entries to configure the costs ?
In json we would have something like:

```json
{
  "customization": {
    "changeFaceCost": 20000,
    "changeHairCost": 10000
     ...
  }
}
```

johmarjac: Yes I like that, it´s more categorized. Good point.

src/Rhisis.Core/Structures/Configuration/WorldConfiguration.cs
```diff
@@ -74,7 +74,7 @@ public class WorldConfiguration : BaseConfiguration
         public uint MailShippingCost { get; set; }
 
         [DataMember(Name = "customization")]
-        public Customization Customization { get; set; }
+        public StyleCustomization Customization { get; set; }
```
Eastrall: I suggest you to add the initialization also after the `{ get; set; }` statement so Newtonsoft.Json can instanciate the property. I don't know why, it doesn't uses the default constructor :/
```csharp
public StyleCustomization Customization { get; set; } = new StyleCustomization();
```

PR #170, opened 2019-01-18
Party System: Join, Leave & Kick

johmarjac:
This PR covers issue #33 and provides functionality to the following features:

- Create Parties
- Join Party
- Leave Party
- Kick Member

If the party leader leaves the party, and there is still 2 or more players, a new party leader is being selected. This is most likely the first player in the party member container.

I havent tested much, so not completely sure if it is bug-free already, but it is a start. Lets discuss these first parts on the system.

Thanks 👍


src/Rhisis.World/Game/Components/PartyMemberContainerComponent.cs
```diff
@@ -0,0 +1,136 @@
...
+        public int MemberCount => Members.Count(x => x != null);
+
+        /// <summary>
+        /// Creates a new <see cref="PartyMemberContainerComponent"/> instance.
+        /// </summary>
+        /// <param name="maxPartyMembers"></param>
+        public PartyMemberContainerComponent(int maxPartyMembers)
+        {
+            MaxPartyMembers = maxPartyMembers;
+            Members = new List<IPlayerEntity>(new IPlayerEntity[maxPartyMembers]);
```
Eastrall: Do we really need to initialize the list of players in the party ? Can't we just make a simple list, add and remove members ?

src/Rhisis.World/Game/Structures/Party.cs
```diff
@@ -0,0 +1,173 @@
...
+        /// <summary>
+        /// Creates a new <see cref="Party"/> instance.
+        /// </summary>
+        /// <param name="id"></param>
+        public Party(int id)
+        {
+            var worldConfiguration = DependencyContainer.Instance.Resolve<WorldConfiguration>();
+
+            Id = id;
+            Name = "Party";
```
Eastrall: Why not moving this to the `PartyConfiguration` settings ? 😄

src/Rhisis.World/Game/Components/PartyMemberContainerComponent.cs
johmarjac: Humm, I guess yes, we won't necessarily need it. I will change that.

src/Rhisis.World/Game/Structures/Party.cs
johmarjac: You mean the whole Party class to the PartyConfigurations.cs?

src/Rhisis.World/Systems/Party/PartySystem.cs
```diff
@@ -0,0 +1,161 @@
...
+
+                leader.Party.Party.AddMember(member);
+
+                foreach (var partyMember in leader.Party.Party.PartyMemberContainer)
+                    WorldPacketFactory.SendAddPartyMember(partyMember, leader.Party.Party, member.PlayerData.Id, leader.Object.Name, member.Object.Name);
+            }
+            else
+            {
+                // Create Party with leader and member
+                if(partyManager.CreateParty(out var party))
```
Eastrall: `CreateParty` will always return `true` according to the method body. In this case, the `out` parameter is useless because it can be returned directly by the method.

Codacy (SonarQube analyzer) says: 

> Passing a parameter by reference, which is what happens when you use the out or ref parameter modifiers, means that the method will receive a pointer to the argument, rather than the argument itself. If the argument was a value type, the method will be able to change the argument's values. If it was a reference type, then the method receives a pointer to a pointer, which is usually not what was intended. Even when it is what was intended, this is the sort of thing that's difficult to get right, and should be used with caution.

In this case I think it is useless to pass the `party` variable as `out` parameter. What do you think ?

src/Rhisis.World/Systems/Party/PartySystem.cs
```diff
@@ -0,0 +1,161 @@
...
+            var worldServer = DependencyContainer.Instance.Resolve<IWorldServer>();
+
+            var leader = worldServer.GetPlayerEntityByCharacterId(e.LeaderId);
+            var member = worldServer.GetPlayerEntityByCharacterId(e.MemberId);
+
+            if (leader == null || member == null)
+                return;
+
+            if (member.Party.IsInParty)
+                return;
```
Eastrall: Can you add a little log here to inform that the player is already in a party? It will help us diagnose errors in case of feedbacks ;)

src/Rhisis.World/Systems/Party/PartySystem.cs
johmarjac: Sure, I will refactor this, I have done it that way because I had another solution before in this method. But now it doesnt make sense to use the out parameter. You are right. I am going to change this as well :+1:

src/Rhisis.World/Systems/Party/PartySystem.cs
johmarjac: Of course :+1:

src/Rhisis.World/Game/Components/PartyMemberContainerComponent.cs
johmarjac: I am going to change this when I am home, I cannot test that critical change at work :+1:

src/Rhisis.World/Game/Structures/Party.cs
```diff
@@ -0,0 +1,174 @@
...
+            packet.Write(Points);
+            packet.Write((int)ExperienceShareMode);
+            packet.Write((int)ItemShareMode);
+            packet.Write(0); // id Duell party?
+
+            for (int i = 0; i < 5; i++) // i < MAX_PARTYMODE wtf?
+                packet.Write(0); // m_nModeTime[i]
+
+            if (IsAdvanced)
+                packet.Write(Name);
```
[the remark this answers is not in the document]
johmarjac: Name of Party can be changed on Party Lv. 10, then party is reserialized to members and also joining members need the party name then.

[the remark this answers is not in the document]
Eastrall: So default party name is set only on client side ?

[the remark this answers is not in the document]
Eastrall: I agree the party name can only be changed when the party reaches level 10. But we can still send a default party name to the client right? And when the user try to change the party name, we check if the party is in advanced mode and the player is the leader of course.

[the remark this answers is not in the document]
johmarjac: It seems so, since the party name is only send if its a bTroup... I think that is the status whether it is an Adv. Party or not. No idea what else it could be. 
Yup, the party when it is initialized the string gets initialized with 0. And the client checks whether the party string is of length 0 and draws the default client sided string, else it draws the string sent from serialization.

image

So we are not able to config it.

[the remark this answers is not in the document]
Eastrall: I see... I guess we can remove the configuration then.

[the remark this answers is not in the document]
johmarjac: Okay, I have corrected this 👍

Eastrall:
Nice PR ! Let's merge

PR #208, opened 2019-04-25
Easier Multiple Database providers and small Rhisis.Database refactor

MarkWilds:
I am mainly using PostgreSQL and wanted this project to have easier mutliple DB provider support.
The Rhisis.Database project is a bit refactored and a little leaner, each migration compatibility set resides in it's own assembly.

**Current migrations assemblies**:
Rhisis.PostgreSQL (PostgreSQL)
Rhisis.MsySQL (MySQL and MsSQL)

**Updating database**:
`cd src/Rhisis.PostgreSQL` (or any other migration assembly)
`dotnet ef database update -s ../Rhisis.World`

Try to see if it works on your configuration.

Updated all other projects to reflect my changes.


MarkWilds:
Still testing and solving errors

MarkWilds:
Fixed most errors and issues I could find.
Would be glad if an MsSQL or MySQL user could test my changes.
I Tested myself, but I am new to this project. Just to be sure.

src/Rhisis.PostgreSQL/Rhisis.PostgreSQL.csproj
```diff
@@ -0,0 +1,11 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+
+    <PropertyGroup>
+        <TargetFramework>netstandard2.0</TargetFramework>
+        <RootNamespace>PostgreSQL</RootNamespace>
```
Eastrall: Maybe it should be: `Rhisis.PostgreSQL` ? what do you think ?

MarkWilds: Yeah, good find, my guess is also I can remove those in the projects and start using assembly name root namespace.

Eastrall: Yes, I guess you can remove it from the `csproj` and use the assembly namespace, which would be something like: `Rhisis.PostgreSQL` by default.
Now that I am thinking about it, shouldn't be better if we name the assemblies : `Rhisis.Database.ProstgreSQL` / `Rhisis.Database.MysSQL` ?

PR #209, opened 2019-04-26
New Flarine NPC dialog: Trinity sister and Kidmen

Anjuts:


bin/data/dialogs/en/MaFl_Is.json
```diff
@@ -0,0 +1,54 @@
...
+			"id": "PRESENTATION",
+			"title": "Who are you?",
+			"texts": [
+				"I am Is, and I am one of the guides in Flarine City. If you need any help or information, just ask me or my twin sisters!",
+				"There are severals of us throughout Flarine and we all look alike so try not to be get confused, tee hee! We confuse ourselves sometimes so it's alright if you do.",
+				"Anyway, stop by and see me anytime if you need any help or information. That's what we're all here for!"
+			 ]
+		},
+		{
+			"id": "PRESENTATION",
```
Eastrall: You might have some issues with duplicate ids. You need to have different and unique dialog link ids. For example, the `PRESENTATION` id, identifies the "Who are you?" dialog, which is a presentation.
For the "Where am I?" you could use a different id like: `LOCATION`, or simply `WHERE_AM_I`.

If you don't do it, it will cause unexpected behaviors, like: when you click on the "Where am I?" link, it will find first texts matching the dialog link id, which in this case is `PRESENTATION", and the first one is "Who are you?" dialog. So it will print you the "Who are you?" dialog instead of the "Where am I?" dialog.

Eastrall: With PR #213 you should have an error on dialogs that have duplicate dialog link keys. It should help you see which dialogs needs a rework. 😄

Eastrall:
Looks good to me now. Let's merge this dialogs and try this out ! 😉

PR #265, opened 2019-09-03
Adding Models for each Configuration File, Some little fixes

MypowerHD:
Iv added few Models for easy handling with the Configuration files

### DIFF
--- a/src/Rhisis.CLI/Commands/Configure/LoginServerConfigurationCommand.cs
+++ b/src/Rhisis.CLI/Commands/Configure/LoginServerConfigurationCommand.cs
@@ -4,6 +4,7 @@ using Rhisis.Core.Helpers;
 using Rhisis.Core.Structures.Configuration;
 using System;
 using System.Collections.Generic;
+using Rhisis.CLI.Models;
 
 namespace Rhisis.CLI.Commands.Configure
 {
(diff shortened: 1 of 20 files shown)


src/Rhisis.CLI/Models/DatabaseConfigurationModel.cs
```diff
@@ -0,0 +1,16 @@
+﻿using Newtonsoft.Json;
+using Rhisis.Database;
+
+namespace Rhisis.CLI.Models
+{
+    public class DatabaseConfigurationModel
```
Eastrall: Not sure that we need this model, since the configuration is using the `DatabaseConfiguration` model from `Rhisis.Database`.

MypowerHD:
Should be fixed 👍

src/Rhisis.CLI/Commands/Configure/LoginServerConfigurationCommand.cs
```diff
@@ -53,17 +56,20 @@ public void OnExecute()
 
             bool response = this._consoleHelper.AskConfirmation("Save this configuration?");
 
-            if (response)
+            if (!response) 
+                return;
+            var configuration = new Dictionary<string, object>
```
Eastrall: It seems that this variable is not used anymore, we can remove it.

MypowerHD: Ohh God :D, I forgot to remove it :D, fixed within next commit

Eastrall:
All good, let's merge this!

PR #278, opened 2019-11-14
Delete Item from Inventory & fixing error when displaying SFX object's at DecreaseItem()

Freezeraid:


src/Rhisis.Network/Packets/World/RemoveInventoryItemPacket.cs
```diff
@@ -34,7 +34,7 @@ public RemoveInventoryItemPacket(INetPacketStream packet)
         /// <param name="other">Other <see cref="RemoveInventoryItemPacket"/></param>
         public bool Equals(RemoveInventoryItemPacket other)
```
Eastrall: You can remove this method now, it will not be used.

PR #279, opened 2019-11-15
Creating a monster by a command 

Freezeraid:


Freezeraid:
Just noticed that creating more than 1 monster in one command makes the mobs not move by themselves after their first move. But they can fight with the player, follow him etc ... Maybe a problem with instancing IMonsterEntity in the ProcessMonsterMovements() function ?

Eastrall:
> Just noticed that creating 2 monster or more in one command makes the mobs buggy. It seems that their behavior are shared and they can't do the same act in the same time, ex :
> 
> * 2 monsters can't attack me in the same time
> * 2 monsters can't choose and go to a destination in the same time
> 
> edit : Indeed, when i let just 1 monster, It begins to act as a normal monster (moving etc ...)

It's actually a weird behavior. When you create a monster using the `IMonsterFactory.CreateMonster()` it internaly requests the monster behavior using the `IBehaviorManager.GetBehavior()` which also internaly, create a new instance of the behavior:

https://github.com/Eastrall/Rhisis/blob/b86fa8d9c201ace96162f8cfe897205270a758c3/src/Rhisis.World/Game/Behaviors/BehaviorManager.cs#L111-L116

If it doesn't find the custom behavior for the `moverId` it will request the default one and create a new instance:

https://github.com/Eastrall/Rhisis/blob/b86fa8d9c201ace96162f8cfe897205270a758c3/src/Rhisis.World/Game/Behaviors/BehaviorManager.cs#L153-L154

Maybe the summoned monster are missing some parameters to attack and behave like other monsters?

Freezeraid:
It's quite weird because, to spawn my mobs, i'm more or less following the spawning monster method of CreateLayer : 
https://github.com/Eastrall/Rhisis/blob/develop/src/Rhisis.World/Game/Maps/MapFactory.cs#L54-L72

And I think there are no missing parameters because spawning monster 1 by 1 with my command spawns fully functionnal mob.

src/Rhisis.World/Game/Chat/Commands/CreateMonsterCommand.cs
```diff
@@ -0,0 +1,70 @@
...
+            {
+                throw new ArgumentException($"Create monster command must have at least one parameter.", nameof(parameters));
+            }
+
+            IMapInstance currentMap = player.Object.CurrentMap;
+            IMapLayer currentMapLayer = currentMap.GetMapLayer(player.Object.LayerId);
+            var currentPosition = new Vector3();
+            currentPosition.X = player.Object.Position.X;
+            currentPosition.Y = player.Object.Position.Y;
+            currentPosition.Z = player.Object.Position.Z;
```
Eastrall: Instead of mapping the X, Y and Z properties, you can directly use the `Vector3.Clone()` method that will duplicate the actual `Vector3` instance into a new one.

src/Rhisis.World/Game/Chat/Commands/CreateMonsterCommand.cs
```diff
@@ -0,0 +1,70 @@
...
+            int sizeOfSpawnArea = 12;
```
Eastrall: Might be better to convert this value into a `const` field in the chat command class.

src/Rhisis.World/Game/Chat/Commands/CreateMonsterCommand.cs
```diff
@@ -0,0 +1,70 @@
...
+
+            IMapInstance currentMap = player.Object.CurrentMap;
+            IMapLayer currentMapLayer = currentMap.GetMapLayer(player.Object.LayerId);
+            var currentPosition = new Vector3();
+            currentPosition.X = player.Object.Position.X;
+            currentPosition.Y = player.Object.Position.Y;
+            currentPosition.Z = player.Object.Position.Z;
+            int sizeOfSpawnArea = 12; 
+            var simulatedRegion = new Rectangle((int)player.Object.Position.X-sizeOfSpawnArea/2, (int)player.Object.Position.Z-sizeOfSpawnArea/2, sizeOfSpawnArea, sizeOfSpawnArea);
+            int monsterId = Convert.ToInt32(parameters[0]);
```
Eastrall: If the first parameter is not an integer formated as a `string`, the `Convert.ToInt32()` method will throw [`FormatException`](https://docs.microsoft.com/en-us/dotnet/api/system.convert.toint32?view=netframework-4.8#System_Convert_ToInt32_System_String_).
I guess you should use the `int.TryParse(stringValue, out int integerValue)` method instead and check the returned value: https://docs.microsoft.com/en-us/dotnet/api/system.int32.tryparse?view=netcore-3.0

src/Rhisis.World/Game/Chat/Commands/CreateMonsterCommand.cs
```diff
@@ -0,0 +1,70 @@
...
+            int quantityToSpawn = parameters.Length >= 2 ? Convert.ToInt32(parameters[1]) : 1;
```
Eastrall: Same remark as above.

src/Rhisis.World/Game/Chat/Commands/CreateMonsterCommand.cs
```diff
@@ -0,0 +1,70 @@
...
+            int sizeOfSpawnArea = 12; 
+            var simulatedRegion = new Rectangle((int)player.Object.Position.X-sizeOfSpawnArea/2, (int)player.Object.Position.Z-sizeOfSpawnArea/2, sizeOfSpawnArea, sizeOfSpawnArea);
+            int monsterId = Convert.ToInt32(parameters[0]);
+            int quantityToSpawn = parameters.Length >= 2 ? Convert.ToInt32(parameters[1]) : 1;   
+
+
+            for (int i = 0; i < quantityToSpawn; i++) {
+                IMonsterEntity monsterToCreate = this._monsterFactory.CreateMonster(currentMap, currentMapLayer, monsterId, currentPosition);
+                monsterToCreate.Object.IsSummoned = true;
+                monsterToCreate.Rectangle = simulatedRegion;
```
Eastrall: Why do you need a `Rectangle` property here ? Instead, you can create a custom `MapRespawnRegion` to will be only assigned to the monster you have summoned.

src/Rhisis.World/Game/Entities/IMonsterEntity.cs
```diff
@@ -14,5 +15,10 @@ public interface IMonsterEntity : IWorldEntity, IMovableEntity, ILivingEntity
         /// Gets or sets the monster's mover data.
         /// </summary>
         MoverData Data { get; set; }
+
+        /// <summary>
+        /// Gets or sets the an area of the spawned monster.
+        /// </summary>
+        Rectangle Rectangle { get; set; }
```
Eastrall: See remark on `CreateMonsterCommand.cs` about the `Rectangle` property.

src/Rhisis.World/Game/Entities/MonsterEntity.cs
```diff
@@ -17,6 +18,9 @@ public class MonsterEntity : WorldEntity, IMonsterEntity
         /// <inheritdoc />
         public IMapRespawnRegion Region { get; set; }
 
+        /// <inheritdoc />
+        public Rectangle Rectangle { get; set; }
```
Eastrall: Same here.

PR #280, opened 2019-11-19
Create Command to spawn monster as an administrator + Change the way to spawn monster in a Map's Layer

Freezeraid:
There is still a bug with spawning multiple monsters with the same command :
- Monsters act like they shared their behavior, they can't move, can't attack at the same time : indeed, if i let just one monster of my command alive, it begins to move and act normally.


src/Rhisis.World/Game/Chat/Commands/CreateMonsterCommand.cs
```diff
@@ -0,0 +1,80 @@
...
+
+        /// <inheritdoc />
+        public void Execute(IPlayerEntity player, object[] parameters)
+        {
+            if (parameters.Length <= 0 || parameters.Length > 2)
+            {
+                throw new ArgumentException($"Create monster command must have 1 or 2 parameters.", nameof(parameters));
+            }
+
+            if (!Int32.TryParse((string)parameters[0], out int monsterId)) {
```
Eastrall: Don't use the .NET types (`Int32`) but prefer using system types: `int` instead. Also, there is always this long debate about the brakets position. C# tends to place the opening braket under the instruction:

```cs
if (condition)
{
    // Do something
}
```

src/Rhisis.World/Game/Factories/Internal/MonsterFactory.cs
```diff
@@ -65,19 +56,59 @@ public IMonsterEntity CreateMonster(IMapInstance currentMap, IMapLayer currentMa
             monster.Moves = new MovableComponent
             {
                 Speed = moverData.Speed / 2,
-                DestinationPosition = monster.Object.Position.Clone()
             };
             monster.Attributes.ResetAttribute(DefineAttributes.STR, moverData.Strength);
             monster.Attributes.ResetAttribute(DefineAttributes.STA, moverData.Stamina);
             monster.Attributes.ResetAttribute(DefineAttributes.DEX, moverData.Dexterity);
             monster.Attributes.ResetAttribute(DefineAttributes.INT, moverData.Intelligence);
-            monster.Behavior = this._behaviorManager.GetBehavior(BehaviorType.Monster, monster, moverId);
```
Eastrall: Why did you removed the behavior from the monster ? if you create a monster, is will not have any behavior and will be static on the world.

PR #283, opened 2019-11-29
Fix multiple teleporting bugs for Visible entities

Freezeraid:


src/Rhisis.World/Packets/Internal/PlayerPacketFactory.cs
```diff
@@ -41,6 +41,7 @@ public void SendPlayerTeleport(IPlayerEntity player)
                 packet.Write(player.Object.Position.Z);
 
                 player.Connection.Send(packet);
+                this._packetFactoryUtilities.SendToVisible(packet, player);
```
Eastrall: Instead of this, you can use the `SendToVisible(packet, player, sendToPlayer)` overload:

```cs
this._packetFactoryUtilities.SendToVisible(packet, player, sendToPlayer: true);
```

src/Rhisis.World/Packets/Internal/PlayerPacketFactory.cs
```diff
@@ -41,7 +41,7 @@ public void SendPlayerTeleport(IPlayerEntity player)
                 packet.Write(player.Object.Position.Z);
 
                 player.Connection.Send(packet);
```
Eastrall: Now that you have set the `sendToPlayer` to `true`, the `SendToVisible` method will send the packet to every visible players including the current player. So, you don't need to send it first. You should delete this line, otherwise, current player will receive the information twice and this can lead to unexpected behaviors.

PR #285, opened 2019-11-30
create item using item name

p1vynodejs:
also added a way to stop InventorySystem.CreateItem if Item is null

### DIFF
--- a/src/Rhisis.World/Game/Chat/Commands/CreateItemCommand.cs
+++ b/src/Rhisis.World/Game/Chat/Commands/CreateItemCommand.cs
@@ -51,7 +51,6 @@ namespace Rhisis.World.Game.Chat
                 this._textPacketFactory.SendDefinedText(player, DefineText.TID_GAME_LACKSPACE);
                 return;
             }
-            //int itemId = Convert.ToInt32(parameters[0]);
             int quantity = parameters.Length >= 2 ? Convert.ToInt32(parameters[1]) : 1;
             byte refine = parameters.Length >= 3 ? Convert.ToByte(parameters[2]) : (byte)0;
             byte element = parameters.Length >= 4 ? Convert.ToByte(parameters[3]) : (byte)0;
--- a/src/Rhisis.World/Game/Chat/Commands/CreateItemCommand.cs
+++ b/src/Rhisis.World/Game/Chat/Commands/CreateItemCommand.cs
@@ -56,11 +56,11 @@ namespace Rhisis.World.Game.Chat
             byte element = parameters.Length >= 4 ? Convert.ToByte(parameters[3]) : (byte)0;
             byte elementRefine = parameters.Length >= 5 ? Convert.ToByte(parameters[4]) : (byte)0;
 
-            string itemStr = Convert.ToString(parameters[0]);
+            string itemInput = parameters[0].ToString();
             Item itemToCreate;
-            if (!Int32.TryParse(itemStr, out int itemId))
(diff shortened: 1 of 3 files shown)


src/Rhisis.World/Game/Chat/Commands/CreateItemCommand.cs
```diff
@@ -50,15 +51,24 @@ public void Execute(IPlayerEntity player, object[] parameters)
...
-
-            int itemId = Convert.ToInt32(parameters[0]);
             int quantity = parameters.Length >= 2 ? Convert.ToInt32(parameters[1]) : 1;
             byte refine = parameters.Length >= 3 ? Convert.ToByte(parameters[2]) : (byte)0;
             byte element = parameters.Length >= 4 ? Convert.ToByte(parameters[3]) : (byte)0;
             byte elementRefine = parameters.Length >= 5 ? Convert.ToByte(parameters[4]) : (byte)0;
-            Item itemToCreate = this._itemFactory.CreateItem(itemId, refine, element, elementRefine, player.PlayerData.Id);
 
-            this._inventorySystem.CreateItem(player, itemToCreate, quantity, player.PlayerData.Id);
+            string itemStr = Convert.ToString(parameters[0]);
```
Eastrall: You don't really need to do a `Convert.ToString()` on the first parameter. The `parameters` is an array of `object`; you can simply do `parameters[0].ToString()` instead.
Also, for naming convention, you should rename the `itemStr` in something more readable; maybe: `itemInput`

src/Rhisis.World/Game/Chat/Commands/CreateItemCommand.cs
```diff
@@ -50,15 +51,24 @@ public void Execute(IPlayerEntity player, object[] parameters)
...
+            Item itemToCreate;
+            if (!Int32.TryParse(itemStr, out int itemId))
```
Eastrall: Please use the system types (`int`) instead of .NET types (`Int32`).

src/Rhisis.World/Game/Factories/Internal/ItemFactory.cs
```diff
@@ -41,21 +42,36 @@ public ItemFactory(ILogger<ItemFactory> logger, IServiceProvider serviceProvider
...
+        public Item CreateItem(string name, byte refine, byte element, byte elementRefine, int creatorId = -1)
+        {
+            var itemByName = this._gameResources.Items.Where(x => x.Value.Name == name);
+            if (itemByName.Count() == 0)
+            {
+                this._logger.LogWarning($"Cannot find item data for item name: '{name}'.");
+                return null;
+            }
+            var item = itemByName.ElementAt(0);
+            this._gameResources.Items.TryGetValue(item.Value.Id, out ItemData itemData);
```
Eastrall: You are actually selecting **ALL** items that has the name you gave as parameter, then you take only the first using `ElementAt(0)`. And then, you query **again** the `_gameResources.Items` dictionnary to retrieve the correct `ItemData`. You are doing unecessary actions here. 

The `item` variable on line 63 is a `KeyValuePair<int, ItemData>`, where `item.Value` contains your `ItemData`. You don't need to request it again using its id.

You could do it in only on line by using the [`FirstOrDefault()`](https://docs.microsoft.com/en-us/dotnet/api/system.linq.enumerable.firstordefault?view=netcore-3.0#System_Linq_Enumerable_FirstOrDefault__1_System_Collections_Generic_IEnumerable___0__System_Func___0_System_Boolean__) method provided by Linq.

PR #287, opened 2019-12-02
Able to use IK2_MAGIC (fireworks) items and starting for all other items

Freezeraid:


src/Rhisis.World/Packets/Internal/SpecialEffectPacketFactory.cs
```diff
@@ -22,23 +22,33 @@ public SpecialEffectPacketFactory(IPacketFactoryUtilities packetFactoryUtilities
...
+        public void SendSpecialEffect(IWorldEntity entity, int specialEffectId, bool sfxNoFollow)
         {
             using (var packet = new FFPacket())
             {
                 packet.StartNewMergedPacket(entity.Id, SnapshotType.CREATESFXOBJ);
                 packet.Write(specialEffectId);
-                packet.Write(0f); // X
-                packet.Write(0f); // Y
-                packet.Write(0f); // Z
+                if (sfxNoFollow == true)
```
Kaev: `if (sfxNoFollow)`
No need to test true == true because true will always be true. :)

src/Rhisis.World/Systems/Inventory/InventoryItemUsage.cs
```diff
@@ -155,12 +155,24 @@ public void UseBlinkwingItem(IPlayerEntity player, Item blinkwing)
...
         }
 
+        public void UseMagicItem(IPlayerEntity player, Item magicItem)
+        {
+            bool noFollowSfx = true;
+            player.Inventory.ItemInUseActionId = player.Delayer.DelayAction(TimeSpan.FromMilliseconds(magicItem.Data.SkillReadyType), () =>
+            {
+                this._specialEffectSystem.SetStateModeBaseMotion(player, StateModeBaseMotion.BASEMOTION_OFF);
+                player.Inventory.ItemInUseActionId = Guid.Empty;
+                this.DecreaseItem(player, magicItem, noFollowSfx);
```
Kaev: You can do it like that so you don't have to create an unneeded variable:
`this.DecreaseItem(player, magicItem, noFollowSfx: true);`

Eastrall: Good catch @Kaev 😉

Issue #296, opened 2020-01-18
Lower capital table names for a better compatibility with all platforms


tekinomikata:
Eastrall found the issue to be related to https://github.com/PomeloFoundation/Pomelo.EntityFrameworkCore.MySql/issues/526 which was indeed the case.

I've posted a suggestion so we would change the EF table attributes to use lower capital naming as well as updating the migrations to reflect the same, so we would attain a better compatibility with other OS and avoid having to use different settings to be able to run under other platforms or users having issues with the same.

https://discordapp.com/channels/294405146300121088/599534604453412865/668136336984637496

Eastrall:
I think this could be a good idea. Can you fix the table names and regenerate the a clean migration from scratch ? I suggest you to delete the "Migration" folder and create a new one named "InitialMigration".

tekinomikata:
@Eastrall Will do!

Eastrall:
Merged with PR #297

PR #297, opened 2020-01-19
#296 Update Table attributes to lower capital

tekinomikata:
Added missing xml doc field for configuration
Deleted all migrations and started it from scratch as requested


Eastrall:
All good. Merging!

Parbil:
In my case the tables are created as it was done in sql server(Camel Case) but later when entity framework generates sqls it gives only lower case and gives object does not exist error.

I had to lower the table name onmodelcreating()

Is there any other configurations to specify them at EF level?

Eastrall:
If you want to use Rhisis with SQL Server, you will need to install the `EF Core SQL Server` provider and regenerate the migration script.
At the moment, Rhisis only supports MySQL providers.

PR #298, opened 2020-01-22
Prevent several code mistakes

Kaev:
This should make Rhisis more stable and fixed some exploits which could be used to crash the server or maybe even add alot of Penya because Int32 arithmetic instead of Int64 arithmetic was used.

### DIFF
--- a/src/Rhisis.Cluster/Client/ClusterClient.cs
+++ b/src/Rhisis.Cluster/Client/ClusterClient.cs
@@ -79,7 +79,7 @@ namespace Rhisis.Cluster.Client
 
             if (this.Socket == null)
             {
-                this._logger.LogTrace("Skip to handle cluster packet from {0}. Reason: client is not connected.", this.Socket.RemoteEndPoint);
+                this._logger.LogTrace("Skip to handle cluster packet from null socket. Reason: client is not connected.");
                 return;
             }
 
--- a/src/Rhisis.Core/Resources/Loaders/ExpTableLoader.cs
+++ b/src/Rhisis.Core/Resources/Loaders/ExpTableLoader.cs
@@ -54,19 +54,6 @@ namespace Rhisis.Core.Resources.Loaders
 
                 var dropLuck = this.LoadDropLuck(dropLuckBlock);
                 var characterExperience = this.LoadCharacterExperience(expCharacterBlock);
-
-                if (dropLuck is null)
-                {
-                    this._logger.LogWarning("Unable to load exp table. Reason: Cannot find drop luck data.");
-                    return;
-                }
-
-                if (characterExperience is null)
-                {
-                    this._logger.LogWarning("Unable to load exp table. Reason: Cannot find character experience data.");
-                    return;
-                }
-
                 var expTableData = new ExpTableData(dropLuck, characterExperience);
                 this._cache.Set(GameResourcesConstants.ExpTables, expTableData);
             }
--- a/src/Rhisis.Core/Resources/Loaders/ExpTableLoader.cs
+++ b/src/Rhisis.Core/Resources/Loaders/ExpTableLoader.cs
@@ -32,16 +32,28 @@ namespace Rhisis.Core.Resources.Loaders
 
             if (!File.Exists(expTablePath))
             {
-                this._logger.LogWarning("Unable to load exp table. Reason: cannot find '{0}' file.", expTablePath);
+                this._logger.LogWarning("Unable to load exp table. Reason: Cannot find '{0}' file.", expTablePath);
                 return;
             }
 
             using (var expTableFile = new IncludeFile(expTablePath, @"([(){}=,;\n\r\t ])"))
             {
-                var expTableData = new ExpTableData(
-                    this.LoadDropLuck(expTableFile.GetBlock("expDropLuck")),
-                    this.LoadCharacterExperience(expTableFile.GetBlock("expCharacter")));
+                var dropLuck = this.LoadDropLuck(expTableFile.GetBlock("expDropLuck"));
+                var characterExperience = this.LoadCharacterExperience(expTableFile.GetBlock("expCharacter"));
 
+                if (dropLuck is null)
+                {
+                    this._logger.LogWarning("Unable to load exp table. Reason: Cannot find drop luck data.");
+                    return;
+                }
+
+                if (characterExperience is null)
+                {
+                    this._logger.LogWarning("Unable to load exp table. Reason: Cannot find character experience data.");
+                    return;
+                }
+
+                var expTableData = new ExpTableData(dropLuck, characterExperience);
                 this._cache.Set(GameResourcesConstants.ExpTables, expTableData);
             }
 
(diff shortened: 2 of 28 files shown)


src/Rhisis.Cluster/Client/ClusterClient.cs
```diff
@@ -79,7 +79,7 @@ public override void HandleMessage(INetPacketStream packet)
 
             if (this.Socket == null)
             {
-                this._logger.LogTrace("Skip to handle cluster packet from {0}. Reason: client is not connected.", this.Socket.RemoteEndPoint);
+                this._logger.LogTrace("Skip to handle cluster packet from null socket. Reason: client is not connected.");
```
Eastrall: Good catch!

src/Rhisis.Core/Resources/Loaders/ExpTableLoader.cs
```diff
@@ -32,16 +32,29 @@ public void Load()
...
+                
+                var expCharacterBlock = expTableFile.GetBlock("expCharacter");
+                if (expCharacterBlock is null)
+                {
+                    this._logger.LogWarning("Unable to load exp table. Reason: Cannot find character experience data.");
+                    return;
+                }
 
+                var dropLuck = this.LoadDropLuck(dropLuckBlock);
+                var characterExperience = this.LoadCharacterExperience(expCharacterBlock);
```
Eastrall: Maybe you could replace the `var` with the actual type, so we can know that type the `LoadDropLuck()` method returns.

src/Rhisis.Core/Resources/Loaders/MoverLoader.cs
```diff
@@ -70,15 +70,23 @@ public void Load()
 
             using (var moversPropExFile = new IncludeFile(propMoverExPath))
             {
-                foreach (Block moverBlock in moversPropExFile.Statements)
+                foreach (var statement in moversPropExFile.Statements)
                 {
+                    var moverBlock = statement as Block;
+                    if (moverBlock is null)
+                        continue;
```
Eastrall: You can use pattern matching here:
```cs
if (!(statement is Block moverBlock))
{
    continue;
}
```

src/Rhisis.Core/Resources/Wld/WldFile.cs
```diff
@@ -38,9 +39,13 @@ private void Read()
 
             while (!reader.EndOfStream)
             {
-                var line = reader.ReadLine().Trim().ToLower();
+                var lineContent = reader.ReadLine();
+                if (lineContent is null)
+                    continue;
+
+                var line = lineContent.Trim().ToLower();
```
Eastrall: In this case, I don't know if it's necessary to use the real type (`string`) for better readability ? we know it's a string, but maybe a beginner doesn't know that ?

src/Rhisis.Network/Packets/World/SfxIdPacket.cs
```diff
@@ -52,7 +52,7 @@ public SfxIdPacket(INetPacketStream packet)
         /// <param name="other">Other <see cref="SfxIdPacket"/></param>
         public bool Equals(SfxIdPacket other)
```
Eastrall: The packet structure will not be used anymore. It was the first version of the packets when we used systems based with event args.
These `struct`s should be converted to `class`es and implement the `IPacketDeserializer` interface.

src/Rhisis.World/Game/Behaviors/DefaultPlayerBehavior.cs
```diff
@@ -75,7 +75,11 @@ public void OnArrived()
         {
             if (this._player.Follow.IsFollowing && this._player.Follow.Target.Type == WorldEntityType.Drop)
             {
-                this.PickUpDroppedItem(this._player.Follow.Target as IItemEntity);
+                var target = this._player.Follow.Target as IItemEntity;
+                if (target is null)
+                    return;
```
Eastrall: Maybe not a good idea to return here. if the target is null because it has already been picked-up (for example, by another member of group), the follow component will not be reseted and then it will keep a reference.

src/Rhisis.World/Packets/Internal/WorldSpawnPacketFactory.cs
```diff
@@ -176,7 +176,7 @@ public void SendPlayerSpawn(IPlayerEntity player)
                 for (var i = 0; i < 6; i++)
                     packet.Write(i);
                 packet.Write<byte>(0); // Base bag item count
-                for (var i = 0; i < 0; i++)
+                for (var i = 0; i < 0; i++) // TODO
```
Eastrall: this `for` loop is very funny tho. 😏

src/Rhisis.Core/Resources/Loaders/ExpTableLoader.cs
Kaev: It's best practice to use var instead of the fixed type. I can still change it if you want though.

src/Rhisis.Core/Resources/Loaders/MoverLoader.cs
Kaev: Good catch, forgot about that!

src/Rhisis.Core/Resources/Wld/WldFile.cs
Kaev: You can always hover your mouse above the variable and your IDE will show you the type the compiler will use. imo we shouldn't break practice to make it easier for beginners.

src/Rhisis.Core/Resources/Loaders/ExpTableLoader.cs
Eastrall: It's more for a readability purpose, because when you actually read it you don't know (just by reading) what it returns.

src/Rhisis.World/Game/Behaviors/DefaultPlayerBehavior.cs
Kaev: Good catch!

src/Rhisis.World/Packets/Internal/WorldSpawnPacketFactory.cs
Kaev: I guessed it is a filler for stuff that it not yet done

src/Rhisis.Core/Resources/Wld/WldFile.cs
Eastrall: > You can always hover your mouse above the variable and your IDE will show you the type the compiler will use.
True, if you are reading the code on an IDE.

> imo we shouldn't break practice to make it easier for beginners.
Naming explicitly the variable wont break any best practices. In fact, it actually recommended to explicitly type a variable when the context is unclear. https://docs.microsoft.com/en-us/dotnet/csharp/programming-guide/inside-a-program/coding-conventions#implicitly-typed-local-variables
And of course, this makes the code more readable.

Eastrall:
Branch squashed and merged into `develop`. Thanks @Kaev for the contribution! 😄

PR #301, opened 2020-01-23
Update rhisis-cli and removed all this keyword

tekinomikata:
Removed all `this` keyword that were unused in the entire solution
Update rhisis-cli needs to use / instead of \ for properly working on linux

### DIFF
--- a/src/Rhisis.Core/Structures/Vector3.cs
+++ b/src/Rhisis.Core/Structures/Vector3.cs
@@ -17,8 +17,8 @@ namespace Rhisis.Core.Structures
         /// </summary>
         public float X
         {
-            get => this._x;
-            set => this._x = value;
+            get => _x;
+            set => _x = value;
         }
 
         /// <summary>
@@ -26,8 +26,8 @@ namespace Rhisis.Core.Structures
         /// </summary>
         public float Y
         {
-            get => this._y;
-            set => this._y = value;
+            get => _y;
+            set => _y = value;
         }
 
         /// <summary>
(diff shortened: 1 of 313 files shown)


src/Rhisis.Core/Structures/Vector3.cs
```diff
@@ -17,37 +17,37 @@ public class Vector3 : IEquatable<Vector3>
         /// </summary>
         public float X
         {
-            get => this._x;
-            set => this._x = value;
+            get => _x;
+            set => _x = value;
```
Eastrall: I really don't remember why I've done that at the begining of the project. Maybe we could replace the X, Y and Z properties defintion with `{ get; set; }` instead of having a private variable ?

src/Rhisis.Network/Packets/PacketHeader.cs
```diff
@@ -14,10 +14,10 @@ public class PacketHeader
...
         public PacketHeader(INetPacketStream packet)
         {
-            this.Header = packet.Read<byte>();
-            this.HashLength = packet.Read<int>();
-            this.Length = packet.Read<int>();
-            this.HashData = packet.Read<int>();
+            Header = packet.Read<byte>();
+            HashLength = packet.Read<int>();
+            Length = packet.Read<int>();
+            HashData = packet.Read<int>();
```
Eastrall: Maybe not used anymore, but we should consider implementing the `IPacketDeserializer` interface in the  `Rhisis.Network\Packets` classes instead of passing the `INetPacketSteam` through the constructor.

PR #318, opened 2020-01-30
Ability to use perin and convert it into gold

Freezeraid:


Eastrall:
Yes, constants are a good option for this kind of problem.
But, I was thinking; wouldn't it be better to add a configuration value for the perin value in the world server configuration file ?

PR #323, opened 2020-02-16
Added Ability to hit all monsters in the visibility range of your player with /aroundkill command

Freezeraid:


Freezeraid:
Sometimes some monsters around are not killed maybe because of the player.Object.Entities list

src/Rhisis.World/Game/Chat/Commands/AroundKillCommand.cs
```diff
@@ -0,0 +1,51 @@
...
+
+        /// <inheritdoc />
+        public void Execute(IPlayerEntity player, object[] parameters)
+        {
+            ObjectMessageType defaultAttackType = ObjectMessageType.OBJMSG_ATK1;
+            float defaultAttackSpeed = 1.0f;
+
+            for (int i = 0; i < player.Object.Entities.Count; i++)
+            {
+                if (player.Object.Entities[i] is IMonsterEntity livingEntity)
```
Eastrall: A monster entity is a living entity, that's for sure, but by naming convention, you should rename the `livingEntity` into `monsterEntity` since its type is `IMonsterEntity`.

src/Rhisis.World/Game/Chat/Commands/AroundKillCommand.cs
```diff
@@ -0,0 +1,51 @@
...
+            ObjectMessageType defaultAttackType = ObjectMessageType.OBJMSG_ATK1;
+            float defaultAttackSpeed = 1.0f;
+
+            for (int i = 0; i < player.Object.Entities.Count; i++)
+            {
+                if (player.Object.Entities[i] is IMonsterEntity livingEntity) 
+                {
+                    if (!livingEntity.Follow.IsFollowing) {
+                        _followSystem.Follow(livingEntity, player);
+                    }
```
Eastrall: The `/aroundkill` command kill every monster around the player right? So you don't need to make the monsters follow the player since they are going to die. 😉

Freezeraid: In fact the original behavior of the command does not kill the monsters around but hit them. It kills them only if you are in onekill command logically. The command is primary misnamed ^^

Eastrall: I see, that's good to know!

Eastrall:
Before merging this PR, please fix the conflicts with the following files:
* `src/Rhisis.World/Systems/Inventory/InventoryItemUsage.cs`
* `src/Rhisis.World/Systems/Inventory/InventorySystem.cs`

This is an issue with your perin commits that shouldn't be in the PR

PR #324, opened 2020-02-17
Ability to summon someone with the /summon command

Freezeraid:
/!\ : Player summoned is not precisely summon on the player if the summoner move before the command. However, teleport or jump before the command ensures that the player summoned is precisely teleported on the summoner.


Freezeraid:
It seems that it could exist a little difference between a player position client side and server side that is refreshed when you jump or you teleport

PR #329, opened 2020-02-22
Ability to send a System message to connected players

Freezeraid:


Freezeraid:
Atm it sends the packet to the visible players only, is there a way to send it to all connected players on the worldserver?

Eastrall:
The `SendToAll()` method is in the `IWorldServer` interface, because it inherits from the `INetServer`.
The `IWorldServer` can be injected in systems and chat commands.

src/Rhisis.World/Game/Chat/Commands/SystemCommand.cs
```diff
@@ -28,11 +28,11 @@ public void Execute(IPlayerEntity player, object[] parameters)
         {
             if (parameters.Length >= 1)
             {
-                string sysMsg = $"[{player.Object.Name}] ";
+                string sysMsg = $"[{player.Object.Name}]";
                 for (int i = 0; i < parameters.Length; i++)
                 {
-                    sysMsg += parameters[i].ToString();
-                }
+                    sysMsg += " " + parameters[i].ToString();
```
Eastrall: I suggest you to avoid using this syntax when you want to concatenate strings. Each time you use the `+=` operator on a `string`, it creates a new instance of a `StringBuilder`. Meaning, if you have a long sentence, let's say 1000 words, it will create 1000 `StringBuilder` instances.

Use the `string.Join()` method instead. It creates internally a `StringBuilder` and appends all the words one after another, just like you have done. But it's better in terms of performances.

PR #330, opened 2020-02-23
Ability to get the numbers of connected players with the /count command

Freezeraid:


src/Rhisis.World/WorldServer.cs
```diff
@@ -107,5 +107,8 @@ public IPlayerEntity GetPlayerEntity(string name)
         /// <inheritdoc />
         public IPlayerEntity GetPlayerEntityByCharacterId(uint id) 
             => Clients.FirstOrDefault(x => x.Player.PlayerData.Id == id)?.Player;
+
+        public uint GetOnlineConnectedPlayerNumber()
```
Eastrall: You are missing the `<inheritdoc />` tag here 😄

PR #352, opened 2020-04-25
351: initial setup fixes

MarkWilds:


MarkWilds:
Best to squash these commits before merging

Issue #350, opened 2020-04-23
Improve ISC communication system


Eastrall:
The FLYFF infrastructure has been built around the a Cluster and Channel system. This means that a single server can have multiple channels.
For example, a server called `Rhisis` can have two channels named `Channel 1` and `Channel 2`.
When a player connects to `Channel 1` it every other players connected to the same channel. 

The cluster has an important role in data centralization of each channels. Like keep a list of connected players and in which channel they are currently playing. Sometimes, when a player makes an action on a channel, the action **MUST** be replicated on the other channels. 
In other words, a channel should be able to broadcast a message to one, multiple or all channels of a cluster.

## The actual system

In the Rhisis context, the **Server** is the `ClusterServer` and the **Channel** is the `WorldServer`.
An Inter-Server Communication system (**ISC**) already exists in the Rhisis project. It is really simple and does the job. But I believe we can do something better.
At the moment, the ISC structure looks like:
![image](https://user-images.githubusercontent.com/4021025/80117504-8fed6380-8587-11ea-81f1-6ffdaeaf17bc.png)

Here, we centralize everything on the `Login Server` that is responsible for handling incoming player authentications. But also has a `Core Server` orignaly made to centralize the server management and broadcast system.
This has been made to centralize the cluster and channel list that will be sent to a new player when it connects to the `Login Server`.
So, when the a `WorldServer` starts, it connects to the `CoreServer` and will appear in the graphic client's server/channel list.

This has a heavy drawback, FLYFF can display many clusters. So when it comes to broadcast messages from a world server to its cluster, it will have to send the message to the `CoreServer` and the broadcast to every channels. If we have X clusters with Y channels and each one of them has N players that send a request to the `CoreServer`, the server will just blow up.

## The improved system idea

The idea of the new version is to centralize every cluster specific actions in the `ClusterServer` project. Just like the `LoginServer` along with the `CoreServer`, the `ClusterServer` will also have a second server running on the same application.

This will result on a the following schema:

![image](https://user-images.githubusercontent.com/4021025/80124750-03e03980-8591-11ea-8045-56a607a6280b.png)

The `WorldServer`s (channels) will now connect to its parent `ClusterServer` instead of connecting the `CoreServer` present on the `LoginServer`.
[...]

PR #355, opened 2020-04-30
#350 - Improve ISC communication system

MarkWilds:
implemented first decoupling as discussed in #350 

- [x] Change initial server infrastructure so that worlds can potentially communicate through cluster without stressing the core server.
- [x] Update Rhisis.CLI to create configuration files based on new server infrastructure.
- [x] Reuse core packet types correctly (probably needs context based action handlers in Sylver.HandlerInvoke.)


MarkWilds:
I would like you to review it when you have time and check if I programmed it the way you envisioned.

src/Rhisis.CLI/Commands/Configure/WorldServerConfigurationCommand.cs
```diff
@@ -54,13 +54,13 @@ public void OnExecute()
...
-            coreConfiguration.Fill();
-            coreConfiguration.Value.Password = MD5.GetMD5Hash(coreConfiguration.Value.Password);
+            Console.WriteLine("----- World cluster Server -----");
+            worldClusterServerConfiguration.Fill();
+            worldClusterServerConfiguration.Value.Password = MD5.GetMD5Hash(worldClusterServerConfiguration.Value.Password);
 
             Console.WriteLine("##### Configuration review #####");
             worldConfiguration.Show("World Server configuration");
-            coreConfiguration.Show("Core server configuration");
+            worldClusterServerConfiguration.Show("World clusterconfiguration");
```
Eastrall, line 63: There is a missing space here. Not a big deal.

src/Rhisis.Cluster/CoreClient/Packets/ICorePacketFactory.cs
```diff
@@ -13,5 +15,12 @@ public interface ICorePacketFactory
         /// <param name="client">Client.</param>
         /// <param name="clusterConfiguration">Cluster server configuration.</param>
         void SendAuthentication(IClusterCoreClient client, ClusterConfiguration clusterConfiguration);
+        
+        /// <summary>
+        /// Sends the world server list to a cluster client.
+        /// </summary>
+        /// <param name="client">Cluster client.</param>
+        /// <param name="worldServers">World server list.</param>
+        void SendUpdateWorldList(IClusterCoreClient client, IEnumerable<WorldServerInfo> worldServers);
```
Eastrall, line 24: When you call the method, you actually send the list of world servers back to the Core Server right ? Might have a little mistake on the documentation summary.

src/Rhisis.Cluster/Handlers/CharacterHandler.cs
```diff
@@ -35,13 +37,16 @@ public class CharacterHandler
...
+        /// <param name="worldCache">World server information cache</param>
+        public CharacterHandler(ILogger<CharacterHandler> logger, IRhisisDatabase database, IClusterServer clusterServer, 
+            IGameResources gameResources, IClusterPacketFactory clusterPacketFactory, IWorldClusterServer worldClusterServer)
         {
             _logger = logger;
             _database = database;
             _clusterServer = clusterServer;
             _gameResources = gameResources;
             _clusterPacketFactory = clusterPacketFactory;
+            _worldClusterServer = worldClusterServer;
```
Eastrall, line 49: I don't think it's a good idea to inject the `WorldClusterServer` in a handler. In terms of responsability, we only need the connected worlds list that is somehow **cached** into the application.
Maybe we could introduce a cache mecanism similar to the `IGameResources`. *It internaly uses the `IMemoryCache` from .NET.*

Furthermore, it could be useful if we want to cache the connected players information such as the name, level, job, gender, etc...

src/Rhisis.World/ClusterClient/Packets/ClusterPacketFactory.cs
```diff
@@ -0,0 +1,28 @@
...
+            using (var packet = new NetPacket())
+            {
+                var config = client.WorldServerConfiguration;
+                packet.Write((uint)CorePacketType.Authenticate);
+                packet.Write(config.Id);
+                packet.Write(config.Name);
+                packet.Write(config.Host);
+                packet.Write(config.Port);
+                packet.Write(config.ClusterId);
+                packet.Write(client.WorldClusterClientConfiguration.Password);
```
Eastrall, line 23: You could use the `config` variable. Also, don't forget to type the variable when it's not explicit.

src/Rhisis.Cluster/Handlers/CharacterHandler.cs
MarkWilds: In terms of responsiblity it doesnt matter that the world server is injected into this handler, BUT it does matter that the world server class handles caching.

I agree on the cache, I had this implemented but removed it because I tried to be consistant with the rest of the code already there :D
image

I will refactor this functionality into a cache looking at IGameResources implementations.

src/Rhisis.World/ClusterClient/Packets/ClusterPacketFactory.cs
MarkWilds: This isnt the same configuration sadly.
What do you mean by "don't forget to type the variable when it's not explicit."?

Eastrall: Oh my bad then. I didn't noticed that there was a  `WorldServerConfiguration` and a `WorldClusterClientConfiguration`.
When I mean "when it's not explicit" its that you don't know the type of your variable. Like in this case, you don't know that type the variaible is:
```csharp
var config = client.WorldServerConfiguration;
```
vs
```csharp
WorldServerConfiguration config = client.WorldServerConfiguration;
```

MarkWilds: I personally believe the right hand side conveys the type for me to justify using var here.
But I understand in this example that its probably a bit vague.

As per C# conventions: 
image

src/Rhisis.Cluster/WorldCache.cs
```diff
@@ -0,0 +1,62 @@
+﻿using System.Collections.Concurrent;
+using System.Collections.Generic;
+using Microsoft.Extensions.Caching.Memory;
+using Rhisis.Core.Resources;
+using Rhisis.Network.Core;
+
+namespace Rhisis.Cluster
+{
+    public class WorldCache : ICache<int, WorldServerInfo>
```
Eastrall, line 9: Ok for this implementation for now. But we will need to have an higher level cache to store, world servers, connected players, etc... in the future.

Eastrall:
All good. Merging into `develop`. Thanks !

Issue #341, opened 2020-04-02
Items database structure improvements


Eastrall:
The actual database model for items and character inventory is pretty simple. Everything is located in a simple table named `Items`. This table contains everything related to the items and inventory. (Slot, Quantity, bonuses, etc...)

This structure has some issues when we start to deal with systems like the `MailSystem` or the `BankSystem` since the `Items` table is only for **inventory items**.

The database structure needs to be changed for a better structure to support all sytems related to items. This is what I propose:
### Remove all non-essential columns from `Items`

`Items` table needs to be more general

### Introduce a general "storage" table that will have the following columns:

| Column Name | Column Type | Comment |
| ----------------|----------------|-------------|
| `CharacterId`   | `INT` | Every item storage will be linked to a character |
| `StorateTypeId` | `INT` | A storage will have a type, that could be : Inventory, Bank or extra bag |
| `ItemId` | `INT` | This is a foreign key to the item that lives in the general `Items` table. |
| `Slot` | `SMALLINT` or `TINYINT` | This is the slot of the item in the storage container. |
| `Quantity` | `SMALLINT` or `INT` | The item quantity. |
| `Updated` | `DATETIME` | The last time the item was updated. (changed slot, increase/decrease quantity, change owner, etc... |
| `Status` | `TINYINT` | For now it will contain the status: `Active` or `Deleted`. But in the future, we could have more statuses. |

### Item attributes

In the future, with the weapon awekening system, items/weapons will have attributes like `+3 DEX` or `-6% Speed`. In order to prepare for such system, we need to add some extra tables to the database to link an item to multiple attributes.

### Introduce an `Attributes` table

This table will contain all the attributes of the `DefineAttributes` enum. We will seed the table using the `Entity Framework Core` `HasValues()` method during a migration.
The table will only contain 2 columns:

| Column Name | Column Type | Comment |
| ----------------|----------------|-------------|
| `Id` | `INT` | The attribute id |
| `Name` | `VARCHAR(20)` | The attribute name |

### Introduce an `ItmeAttributes` table

This table will be used as a many to many table between items and attributes.
One Item can have multiple attributes.
One attribute can be linked to multiple items but with different bonuses.

| Column Name | Column Type | Comment |
| ----------------|----------------|-------------|
| `ItemId` | `INT` | Foreign key to the `Items` table and `Id` column. |
| `AttributeId` | `INT` | Foreign key to the `Attributes` table and `Id` column. |
[...]

PR #416, opened 2020-05-24
Update item database structure

Eastrall:
This PR covers issue #341 and changes the item database schema.

It introduces the following changes:

* Add `Attributes` table which contains the list of all attributes based on the `DefineAttributes` enum.
* Add `InventoryItems` table which contains the actual item position and quantity that the player possesses.
* Add `ItemAttributes` table which contains an item attributes such as +4 INT or -6% speed. *This will be useful for the awekening system*
* Update `Item` table. This table is now more general and only contains the `ItemId`, `Refine (nullable)`, `Element (nullable)` and `ElementRefine (nullable)`. This table has a many-to-many to the `Attributes` table using the `ItemAttributes` join table.

> ⚠️ You will need to update your database schema. The EF migration migrates the previous `Items` table to the `InventoryItems` table and seeds the `Attributes` table with the `DefineAttributes` enum values.

Also, the previous `Item` structure has been renamed into `InventoryItem` which now represents an inventory item. More structures like `BankItem`, `MailItem` will be created in the future. 

Below the new items database schema with the changes:
test


Eastrall:
Closing this PR. I came up with a better solution and database structure. I'll explain everything in #341

PR #472, opened 2020-11-12
World Server refactoring

Eastrall:
This PR contains a huge rework of the WorldServer structure.
In previous version, every entity, factory, system and even packets were tied up with the WorldServer. The goal of this rework is to split up the different parts of it and have a more maintenable environment.

# Purpose

The purpose of this rework is mainly to rearrange the project to match a more "Domain Driven" architecture. We will have a project that contains only abstractions and another one which will contain the implementations.

In the end, the goal is to provide a simple API to develop new systems/features and also to facilitate scripting (Quests, events, IA, etc...).

For example, before, if we wanted to give some experience to a player, we would have to do something like:
```csharp
// Let's assume we are in a handler and the client sends a GIVEEXP packet

public class ExperienceHandler
{
    private readonly IExperienceSystem _experienceSystem;

    // Dependency injection because we are cool dudes.
    public ExperienceHandler(IExperienceSystem experienceSystem)
    {
         _experienceSystem = experienceSystem;
    }

    [HandlerAction(PacketType.GIVEEXP)]
    public void OnGiveExperience(IPlayer player, GiveExperiencePacket packet) // this packet contains the total amount of exp we will give
    {
        _experienceSystem.GiveExperience(player, packet.ExpAmount);
    }
}
```
Here we have some nice code with dependency injection in the handler. We call the system's `GiveExperience()` method and give to it the player and the amount of experience.
This solution was working well so far, when we had a few systems. But in the real world, we ended up with systems that would require a dependency, and this dependency would require the calling system instance. In other words, we would end up with a circular dependency :boom:
This is bad.

So I have studied to the topic and came across and article that talked about mixing the SOA (Service Oriented Architecture, what we had) and some traditionnal OOP (Oriented Object Programming). After couple of days of study, I decided to give it a try and create an easy and understandable OOP interface to call my systems. This is where the idea of this refactoring came from.

Now, let's take the example above and use it with the current implementation:
```csharp
// Let's assume we are in a handler and the client sends a GIVEEXP packet

public class ExperienceHandler
{
    // No more dependency injection, but we still love it

    [HandlerAction(PacketType.GIVEEXP)]
    public void OnGiveExperience(IPlayer player, GiveExperiencePacket packet) // this packet contains the total amount of exp we will give
    {
        player.Experience.Increase(packet.ExpAmount);
    }
}
```
Pretty simple right and straightforward right? Our `IPlayer` interface contains a lot of properties and **Features** (experience, gold, Attributes, Skills, Buffs, etc...) that can be called in a oriented object way.
This structure will facilitate the use of quest and event scripting, because we will simply pass the `IPlayer` object to our script method and you will be able to access any player's property and features.

```lua
-- Let's say we are inside a quest script
QUEST_HERO = {
	title = 'IDS_PROPQUEST_INC_002064',
	character = 'MaDa_RedRobeGirl',
	end_character = 'MaDa_RedRobeGirl',
        ...
}

--
-- Returns the job reward based on the player's job.
--
function QUEST_HERO:change_job_reward(player)
	local hero_jobs = {
		[Jobs.JOB_KNIGHT_MASTER] = Jobs.JOB_KNIGHT_HERO,
		[Jobs.JOB_BLADE_MASTER] = Jobs.JOB_BLADE_HERO,
		[Jobs.JOB_BILLPOSTER_MASTER] = Jobs.JOB_BILLPOSTER_HERO,
		[Jobs.JOB_RINGMASTER_MASTER] = Jobs.JOB_RINGHERO_HERO,
		[Jobs.JOB_JESTER_MASTER] = Jobs.JOB_JESTER_HERO,
		[Jobs.JOB_RANGER_MASTER] = Jobs.JOB_RANGER_HERO,
		[Jobs.JOB_PSYCHIKEEPER_MASTER] = Jobs.JOB_PSYCHIKEEPER_HERO,
		[Jobs.JOB_ELEMENTOR_MASTER] = Jobs.JOB_ELEMENTOR_HERO,
	}

        -- Here we access the player job id.
	return hero_jobs[player.Job.Id] or -1
end
```
This opens so much scripting possibilities that the official files cannot really do at the moment.
We could imagine some more hooks like:

```lua
QUEST_HERO = {
    ...
}

function QUEST_HERO:on_finish(player) {
    player.Gold.Increase(100000); -- Give some 100.000 gold to the player
    player.Experience.Increase(500); -- Give some 500 exp to the player
    player.Statistics.Points += 10; -- Give 10 more stat points
}
```

> Note: A list of the available properties and methods for scripting will be done in the future.

# New project structure

The Rhisis solution now includes 3 new projects.

### `Rhisis.Game.Common`

 This project contains all game resources common to every projects, such as the well known `DefineAttributes` enumeration which contains all the game attributes (`STA`, `STR`, `DEX`, `INT`, etc.)

### `Rhisis.Game.Abstractions`

This project contains all the game abstractions. Such as the `IPlayer` that represents a player, but also all the feature abstractions such as the `IGold` feature which manages the player's gold amount; `IExperience` which manages the player's experience and level up process.
You will also find all independent or shared systems abstractions such as the `IQuestSystem` which is considered as a "Shared System".

### `Rhisis.Game`

This is the central project where the game logic is located. All the abstractions contained in the `Rhisis.Game.Abstractions` are implemented in this project.

# Network protocole rework

Previously, all packet structures were stored in "packet factories". We would have a method that will build the packet/snapshot and then send it to a player or a set of players.
The FLYFF protocole provides the possibility to merge snapshots together and send them using only one `Socket.Send()` call back to the client. With the actual system, we had a really neat separation of each packets, but we couldn't merge them together.

In this new version, all packet/snapshot structures are stored in individual classes and can now me merged together:
> Snippet from the `Health` system.
```csharp
private void SendHealth()
{
    using var healthSnapshot = new FFSnapshot();
    healthSnapshot.Merge(new UpdateParamPointSnapshot(_mover, DefineAttributes.HP, Hp));
    healthSnapshot.Merge(new UpdateParamPointSnapshot(_mover, DefineAttributes.MP, Mp));
    healthSnapshot.Merge(new UpdateParamPointSnapshot(_mover, DefineAttributes.FP, Fp));

    SendPacketToVisible(_mover, healthSnapshot, sendToPlayer: true);
}
```
Here, we create a new empty snapshot named `healthSnapshot`. Then we merge an `UpdateParamPointSnapshot` that will contain the new HP value, and we do the same with the MP and FP components.
Finally, we send the `healthSnapshot` back to the client and visible players. In other terms, we send 3 information updates in only one socket `Send()` operation instead of sending 3 individual packets.
This is a network optimization that could be really useful if the server has a lot of connected players.

We can find all the packet structures so far in the `Rhisis.Game/Protocole` folder.


src/Rhisis.Game.Abstractions/IQuest.cs
```diff
@@ -0,0 +1,65 @@
...
+        /// </summary>
+        QuestStateType State { get; set; }
+
+        /// <summary>
+        /// Gets the quest start time.
+        /// </summary>
+        DateTime StartTime { get; }
+
+        /// <summary>
+        /// Gets a dictionnary of the killed monsters. Key is the monster id; Value is the killed amount.
```
SvenLlama, line 56: dictionary

Eastrall: Oops, typo error. I'm going to add more documentation to the abstractions and I'll fix this typo.

Issue #473, opened 2020-11-13
Crash of world server during trace log


SvenLlama:
Exception:
`[2020-11-13 21:34:29.1405] [INFO] New client connected from 127.0.0.1:54103.
Unhandled exception. System.AggregateException: An error occurred while writing to logger(s). (Invalid format string. Expected 4 format parameters, but failed to lookup parameter index 3)
 ---> System.FormatException: Invalid format string. Expected 4 format parameters, but failed to lookup parameter index 3
 ---> System.IndexOutOfRangeException: Index was outside the bounds of the array.
   at Microsoft.Extensions.Logging.LogValuesFormatter.GetValue(Object[] values, Int32 index)
   at Microsoft.Extensions.Logging.FormattedLogValues.get_Item(Int32 index)
   at NLog.Extensions.Logging.NLogMessageParameterList.TryGetParameterName(IReadOnlyList`1 parameterList, Int32 i, Nullable`1& originalMessageIndex, String& parameterKey)
   --- End of inner exception stack trace ---
   at NLog.Extensions.Logging.NLogMessageParameterList.TryGetParameterName(IReadOnlyList`1 parameterList, Int32 i, Nullable`1& originalMessageIndex, String& parameterKey)
   at NLog.Extensions.Logging.NLogMessageParameterList.IsValidParameterList(IReadOnlyList`1 parameterList, Nullable`1& originalMessageIndex, Boolean& hasMessageTemplateCapture, Boolean& isMixedPositional, Boolean& isPositional)
   at NLog.Extensions.Logging.NLogMessageParameterList..ctor(IReadOnlyList`1 parameterList)
   at NLog.Extensions.Logging.NLogMessageParameterList.TryParse(IReadOnlyList`1 parameterList)
   at NLog.Extensions.Logging.NLogLogger.TryParseMessageParameterList(IReadOnlyList`1 messageProperties)
   at NLog.Extensions.Logging.NLogLogger.TryParseMessageTemplate(LogLevel nLogLogLevel, IReadOnlyList`1 messageProperties, NLogMessageParameterList& messageParameters)
   at NLog.Extensions.Logging.NLogLogger.CreateLogEventInfo[TState](LogLevel nLogLogLevel, TState state, Exception exception, Func`3 formatter)
   at NLog.Extensions.Logging.NLogLogger.Log[TState](LogLevel logLevel, EventId eventId, TState state, Exception exception, Func`3 formatter)
[...]

PR #474, opened 2020-11-13
Fix world server crash on unimplemented packets

SvenLlama:


SvenLlama:
Closes #473

Eastrall:
Looks like I've missed to `$` to do string interpolation. My bad. Thanks for the fix! Merging.

PR #475, opened 2020-11-13
Support ONEKILL_MODE in the battle system feature

SvenLlama:
All damage caused by the Battle systems (spells, melee attacks) should be impacted.
I didn't check if official FlyFF also worked like this or if it only triggered on melee attacks, but I think it makes sense like this.

### DIFF
--- a/src/Rhisis.Game/Features/AttackArbiters/MeleeAttackArbiter.cs
+++ b/src/Rhisis.Game/Features/AttackArbiters/MeleeAttackArbiter.cs
@@ -88,11 +88,6 @@ namespace Rhisis.Game.Features.AttackArbiters
         /// <returns></returns>
         private AttackFlags GetAttackFlags()
         {
-            if (Attacker is IPlayer player && player.Mode.HasFlag(ModeType.ONEKILL_MODE))
-            {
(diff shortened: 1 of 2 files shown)


Eastrall:
In the implementation you suggest, you check if the player has the `ONEKILL` mode activated when inflicting the damages, after processing the `AttackArbiter` and `AttackReducer`.
Shouldn't this check be done before any attack calculation? We will have a small code duplication for the `MeleeAttack()` and `SkillAttack()` methods, but at least we don't waste CPU time doing calculation that wouldn't be needed.

SvenLlama:
Oke, I'll change the check to be before the arbiters.
My reasoning was that a singular check was preferred over double checks since it's unlikely for a lot of players to constantly be using this mode on a server and so the additional arbiters before the check wouldn't be noticeable at all.

Eastrall:
You are correct. But we don't know if some servers would want to create a skill that gives the `ONEKILL` mode for like 5s or even a special buff given for a quest. So maybe 2 checks is better to avoid extra CPU usage and unnecessary calculation.
Anyway, looks fine to me now. Merging. Thanks 😄

PR #478, opened 2020-11-14
Add magic skill projectile

Eastrall:
This PR adds the magic skill projectiles support and improves the projectile system.

Before, we used the projectile id coming from the client. This is a security issue. Now each movers that can send projectiles keep a track of their projectile id. Also, when  a player is teleported to another map, the projectile id is restored to its default value : 1.


SvenLlama (approved):

SvenLlama:
Maybe a small thought for later (or now?):
To integrate systems with each other, would we want to have the teleport system call the projectiles system like it was implemented now,
Or would the projectiles systems subscribe on a  teleport event of the teleport system, and execute the reset itself when needed. This should reduce the hard dependencies of the systems on each other.
For this use case it's probably over kill, but for other integrations using a event driven approach could be useful.

Eastrall:
I like the idea of having systems that subscribe to other actions. But this would require some extra work to implement a "reactive" system where other systems/feature could subscribe to event.
Let's keep that in mind for the future. 😉

PR #480, opened 2020-11-15
Refactor Battle feature to no longer be linked with the object messag…

SvenLlama:
…e type bloat

I refactored the Battle feature a little bit so the ObjectMessageType is no longer used directly.
Using the ObjectMessageType in the domain logic makes the code a lot less readable as you don't know what is in there.
Now there is a dedicated enum type AttackType, and the ObjectMessageType is now only used in the packets and snapshots.

There is also a extension class for the AttackType enum which links certain behavior with the types of the enum indirectly, to make the code checking on the enum more readable and more robust for the future.

An alternative for this AttackType + Extensions could be to create a "Enum style class" which would then have properties to replace the extension methods, but I chose to keep it more simple for now.

### DIFF
--- a/src/Rhisis.Game.Abstractions/Features/Battle/RangeAttackTypeExtensions.cs
+++ b/src/Rhisis.Game.Abstractions/Features/Battle/RangeAttackTypeExtensions.cs
@@ -5,6 +5,13 @@ namespace Rhisis.Game.Abstractions.Features.Battle
 {
     public static class RangeAttackTypeExtensions
     {
+        /// <summary>
+        /// The client uses the <see cref="ObjectMessageType"/> enum to indicate the attack type.
+        /// This method will convert the <see cref="AttackType"/> known at the server to the attack type known by the client (<see cref="ObjectMessageType"/>).
+        /// </summary>
+        /// <param name="attackType">The attack type.</param>
+        /// <exception cref="NotImplementedException">The given attack type does not have a matching <see cref="ObjectMessageType"/>.</exception>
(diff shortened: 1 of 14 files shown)


src/Rhisis.Game/Systems/SkillSystem.cs
```diff
@@ -246,7 +246,7 @@ private void ExecuteSkill(IMover caster, IMover target, ISkill skill, bool reduc
                 throw new InvalidOperationException("Only a player or monster can cast a skill.");
             }
 
-            battle.SkillAttack(target, skill);
+            battle.TrySkillAttack(target, skill);
```
Eastrall, line 249: Maybe we should check the return value of this and only decrease points once the skill attack has been executed.

src/Rhisis.Game.Abstractions/Features/Battle/RangeAttackTypeExtensions.cs
```diff
@@ -0,0 +1,103 @@
+﻿using Rhisis.Game.Common;
+using System;
+
+namespace Rhisis.Game.Abstractions.Features.Battle
+{
+    public static class RangeAttackTypeExtensions
+    {
+        public static ObjectMessageType ToObjectMessageType(this AttackType attackType)
```
Eastrall, line 8: Can you provide some summaries documentation here please?

Eastrall (commented):
Nice refactoring. Just some feedbacks about checks and docs, and we're good to merge.

Eastrall (approved):
All good.

PR #489, opened 2020-11-28
Replace ISC with Redis and RabbitMQ. Add CoreServer for server cache

### LINKED ISSUE #488: Replace CoreServer with a messaging queue system
In the actual solution, the servers communicate using a custom socket server on `LoginServer` and `ClusterServer`. These servers are used to broadcast messages to the connected `WorldServers`. Instead of maintaining a custom solution, I believe we should use an existing message queue that (in my opinion) suits perfectly what we want to achieve here.

I propose to use RabbitMQ which is an open-source message queue system.

When PR #487 will be merged, I will create a new docker container with this message queue server and provide an easy API to publish and subscribe to messages instead of using RabbitMQ native methods. This allows us to change the messaging queue system whenever we want without breaking any functionnal code.

Eastrall:
This PR adds Redis and RabbitMQ in replacement with the ISC.

Also, I've added a `CoreServer` that centralizes the Cluster and World channel connections. This is mostly used to update the server's cache list in case of any disconnection.

Bellow a schema of the current architecture
image

> Note that the database part is not implemented yet. And for now, the database is a single database with everything inside.

Closes #488


Eastrall:
After some thoughts, I realized that the use of a messaging queue like `RabbitMQ` might be a little overkill for the use of the project. So, I'm coming back to more simple architecture with an embeded `CoreServer` inside of the `LoginServer` program.

Why this change again? The purpose of the messaging queue was to send messages to every servers connected to it. Which is great to broadcast messages. But this broadcast mechanism is only needed for the Cluster and Channels, so we could use the publish/subscribe mechanism of Redis instead of having an extra server in our infrastructure. And if you think about it, the Clusters and Channels are connected to the `CoreServer` and will only send data like their information when they are alive, pings, and when a player connects/disconnects (to keep a track of the number of players connected on the channels/clusters).

That said, bellow you'll find the new schema of the "typical" infrastructure.

image

PR #507, opened 2023-04-25
Add sit down motion

william-dechamps:


src/Rhisis.WorldServer/Handlers/MotionHandler.cs
```diff
@@ -0,0 +1,28 @@
...
+    private readonly ILogger<MotionHandler> _logger;
+
+    /// <summary>
+    /// Creates a new <see cref="MotionHandler"/> instance.
+    /// </summary>
+    /// <param name="logger">Logger.</param>
+    public MotionHandler(ILogger<MotionHandler> logger)
+    {
+        _logger = logger;
+    }
```
Eastrall, line 21: No need to inject a logger if you don't need it

src/Rhisis.Game/Protocol/Packets/World/Client/MotionPacket.cs
```diff
@@ -0,0 +1,17 @@
...
+using Rhisis.Protocol;
+
+namespace Rhisis.Game.Protocol.Packets.World.Client;
+
+public class MotionPacket
+{
+    /// <summary>
+    /// Gets the motion.
+    /// </summary>
+    public ObjectMessageType Motion { get; private set; }
```
Eastrall, line 11: You can remove the `private set` since you initialize the property in the constructor.
Having a `get` only property is considered as a "readonly property"

PR #514, opened 2023-05-10
Add player taskbar

william-dechamps:


src/Rhisis.Game/Entities/Player.cs
```diff
@@ -99,6 +100,11 @@ public sealed class Player : Mover
     /// </summary>
     public string CurrentShopName { get; set; }
 
+    /// <summary>
+    /// Gets the player taskbar.
+    /// </summary>
+    public Taskbar Taskbar { get; set; }
```
Eastrall, line 106: You can remove the `set` attribute since it's a property that will be initialzed once in the constructor of the entity.

src/Rhisis.Game/Taskbar/Taskbar.cs
```diff
@@ -0,0 +1,29 @@
...
+        Applets = new TaskbarContainer<Shortcut>(18);
+        Items = new MultipleTaskbarContainer<Shortcut>(8, 9);
+    }
+
+    public void Serialize(FFPacket packet)
+    {
+        Applets.Serialize(packet);
+        Items.Serialize(packet);
+        packet.WriteInt32(0);
+        ActionSlot.Serialize(packet);
```
Eastrall, line 26: Before uncomenting the `ActionSlot.Serialize(packet)` line, the `packet.WriteInt32(0)` was here to indicate that there is 0 skills in the action slot. Since you have uncomented the line, you need to remove the `packet.WriteInt32()` line since it's already in the `Serialize()` method.

Eastrall (approved):

PR #515, opened 2023-05-10
Add GM command /stat

william-dechamps:


src/Rhisis.Game/Chat/Commands/StatCommand.cs
```diff
@@ -0,0 +1,67 @@
...
+            default:
+                break;
+        }
+
+        _logger.LogTrace($"Player {player.Name} update attribute {attribute} value to {quantity}");
+
+        using (var snapshot = new ModifyStateSnapshot(player))
+        {
+            player.Send(snapshot);
+            player.SendToVisible(snapshot, true);
```
Eastrall, line 64: You are sending the packet twice here.
The second parameter of `SendToVisible()` is a boolean that indicates if you also want to send the packet to the current entity. In our case, the player.
Therefor, we only need to call the `SendToVisible(snapshot, sendToSelf: true);`

Eastrall (approved):

